PR #680, opened 2024-02-19
Add docs for `step.invoke()` with `timeout`

jpwilliams:
## Summary

Adds documentation for `timeout` in `step.invoke()` as added in inngest/inngest#1129 and inngest/inngest-js#484.

## Related

- inngest/inngest#1129
- inngest/inngest-js#484


sylwiavargas (approved):
Thank you!

PR #696, opened 2024-02-23
Use branch current build to build edit this page URL

djfarrelly:
This is only mean to make links for newly created MDX documents work on branch builds/preview builds. This also is nice as if someone shares a preview build, the button goes to the correct pull request branch on Github. By default this will be "main."

### DIFF
--- a/shared/Docs/Layout.tsx
+++ b/shared/Docs/Layout.tsx
@@ -15,7 +15,8 @@ import { SectionProvider } from "./SectionProvider";
 import { useMobileNavigationStore } from "./MobileNavigation";
 import { getOpenGraphImageURL } from "../../utils/social";
 
-const GITHUB_PREFIX = "https://github.com/inngest/website/tree/main/";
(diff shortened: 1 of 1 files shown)


shared/Docs/Layout.tsx
```diff
@@ -15,7 +15,8 @@ import { SectionProvider } from "./SectionProvider";
 import { useMobileNavigationStore } from "./MobileNavigation";
 import { getOpenGraphImageURL } from "../../utils/social";
 
-const GITHUB_PREFIX = "https://github.com/inngest/website/tree/main/";
+const GITHUB_BRANCH = process.env.NEXT_PUBLIC_VERCEL_GIT_COMMIT_REF || "main";
```
djfarrelly, line 18: Tested on Vercel preview - this works correctly on the branch build. We'll have to test after this PR is merged that this is correct on production. `NEXT_PUBLIC_VERCEL_GIT_COMMIT_REF` is supposed to be set to `main` on production, but we should confirm this when out.

sylwiavargas (approved):
Thank you!

PR #702, opened 2024-02-28
Add a note that `onSendEvent` middleware will run with `step.invoke()`

jpwilliams:
## Summary

Adds a note to the TS middleware docs that `onSendEvent()` will be run for calls to `step.invoke()`, following the bug fix in inngest/inngest-js#503.

## Related

- inngest/inngest-js#503


sylwiavargas (approved):
Thank you!

PR #704, opened 2024-02-28
Remove sync dedupe text

goodoldneon:
We no longer dedupe syncs


anafilipadealmeida:
Screenshot 2024-02-28 at 22 44 05
I think there's another callout that needs to be removed

sylwiavargas (changes requested):
Great, thank you!
If you could just delete [this callout](https://github.com/inngest/website/pull/704#issuecomment-1970045536), it's ready to go!

PR #707, opened 2024-03-07
Remove our dashboard from the cancel docs description

goodoldneon:
There isn't a way to cancel function runs from our UI yet, so we shouldn't mention the dashboard in the cancel page's description


sylwiavargas (approved):
Ha, good catch, thank you!

PR #708, opened 2024-03-11
INN-2866 Add multi-trigger reference docs

jpwilliams:
## Summary

Adds small tweaks to Python and TypeScript reference documentation to add a mention of multiple triggers.

Golang uses auto-generated documentation from [pkg.go.dev](https://pkg.go.dev/github.com/inngest/inngestgo).

## Related

- INN-2866
- Documents inngest/inngest-js#497
- Documents inngest/inngest-py#59

PR #715, opened 2024-03-13
A small guide for using multiple triggers

jpwilliams:
## Summary

Adds a small guide for using multiple triggers - something to refer folks to following the release of the feature.

[New page - Using Multiple Triggers](https://website-git-docs-multiple-triggers-guide-inngest.vercel.app/docs/guides/multiple-triggers)

### DIFF
--- a/pages/docs/guides/multiple-triggers.mdx
+++ b/pages/docs/guides/multiple-triggers.mdx
@@ -1,0 +1,94 @@
+# Multiple triggers
+
+Each Inngest function can be triggered by one or more events or schedules, up to
+a maximum of 10.
+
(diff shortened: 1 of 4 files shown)


djfarrelly (approved):
We should change the TS reference docs to also show that multiple triggers are allowed and link back to this guide.

jpwilliams:
> We should change the TS reference docs to also show that multiple triggers are allowed and link back to this guide.

Added a note to the sections added in #708. :)

jpwilliams:
Will ship this without wildcard docs, as they need more DX consideration. This is tracked in DEV-45.

PR #716, opened 2024-03-13
Update step-send-event.mdx

wtachau:
reliability -> reliably


sylwiavargas (approved):
Ohhhh good catch! Thank you 🙏💜

PR #718, opened 2024-03-14
DEV-52 Make the main sidebar an accordion

sylwiavargas:
## Overview

This PR transforms the sidebar menu into a (nested) accordion to save space when there will be more items on the side.

## Considerations
- Only the current "section" of the sidebar is open by default
- If the user opens a "section", the state will persist as the user browses around
  - it will not persist if the page is refreshed (this can be done with session storage but I'm not sure if it's the desired behavior yet)
- The sidebar can start with just a page or a "section"
- The maximum level of nested-ness is two (a subsection within section -- see second screenshot) and having drawn out possible menu items, there doesn't seem to be a need for further nesting
- Each of the main menu section will have a landing page (this does not apply to sub-sections)

## Next steps

- The current implementation does not feature the styling to indicate the active page beyond a vertical line. In the first screenshot, we are on the landing page and no subpage is chosen (this will be different once landing pages are also present in the sidebar). In the second screenshot, the current page is marked

### Sidebar
Screenshot 2024-03-14 at 2 00 52 PM

### Nested sidebar
Screenshot 2024-03-14 at 2 00 40 PM


shared/Docs/navigationStructure.ts
```diff
@@ -158,19 +163,26 @@ const sectionGuides = [
...
       {
         title: "Working with apps",
         href: `/docs/apps/cloud`,
       },
-      { title: "Deploy: Vercel", href: `/docs/deploy/vercel` },
-      { title: "Deploy: Netlify", href: `/docs/deploy/netlify` },
       {
-        title: "Deploy: Cloudflare Pages",
-        href: `/docs/deploy/cloudflare`,
+        title: "Deployment targets",
```
djfarrelly, line 176: Maybe "platforms" instead of targets?

styles/globals.css
```diff
@@ -770,3 +770,45 @@ ul.check li {
 .hero-text-shadow {
   text-shadow: 0 5px 30px rgba(224,198,210,0.2);
 }
+
```
djfarrelly, line 773: **non-blocking**: Ideally, we'd move all of these our of global.css and use tailwind as we use elsewhere in the site. Currently `global.css` is mostly legacy CSS that we're working to shift away from over time. We don't need to do this for this PR, but we should move it at some point for consistency.

djfarrelly (approved):
A couple non-blocking comments, but otherwise, good to go as this will be great to pave the way for the coming IA changes!

styles/globals.css
sylwiavargas: Thank you, this make sense. I've created a [Linear](https://linear.app/inngest/issue/DEV-81/move-code-out-of-globalscss) ticket to get back to this.

djfarrelly (approved):

shared/Docs/navigationStructure.ts
```diff
@@ -477,6 +559,293 @@ const sectionPythonReference = [
...
+  },
+];
+
+const matchers = {
+  guides: /^\/docs\/guides/,
+  reference: /^\/docs\/reference/,
+  examples: /^\/docs\/examples/,
+  // should match everything except above
+  default: /^\/docs(?!\/guides|\/reference|\/examples)/,
+};
```
sylwiavargas, line 820: @djfarrelly matchers

PR #725, opened 2024-03-18
DEV-53 Improved Nav Bar

sylwiavargas:
## Overview

➡️ [Preview](https://website-git-sylwia-tabs-inngest.vercel.app/docs)

This PR introduces the following changes:
- refurbished Nav Bar (in accordance with the feedback from our designers)
- improved mobile navigation (in accordance with the feedback from our designers)
- tabs in the Nav Bar
- [reference landing page](https://website-git-sylwia-tabs-inngest.vercel.app/docs/reference) (very minimal)
- enabling all the sidebar sections to be opened at the same time (something which will change)
- moving all reference sites to one place

## Code quality

There is a lot of debris because it is an (incremental) work in progress. Please be understanding.

---
## Outdated: RFC

Previously, this PR was an RFC and I'm keeping the contents for posteriority.

<details>
<summary>Content from RFC</summary>
## Request for comments

This is a work-in-progress PR to request comments on how to best approach the cluttered Nav Bar -- both in desktop and mobile version.

## The problem

I have added four "Tabs" to the docs nav bar, which makes it a bit cluttered:

### Light mode
Screenshot 2024-03-18 at 10 34 57 AM

### Dark mode
Screenshot 2024-03-18 at 10 34 54 AM

Before I move onto making further decisions or introduce more styling changes (reducing margins,  and so on), I decided to stop to brainstorm best path forward.

## Solutions - desktop
I brainstormed some solutions, looked at available patterns elsewhere and compiled these ideas into a list below. Please note that the mockups are made in Preview so they can be a bit uneven etc. Also, maybe we can come up with something else altogether or combine a few solutions -- I'm open to suggestions.

### 1. Social icons in the sidebar
We could move the social icons to the bottom of the sidebar, and then move search box to the right. The con here is that the social icons are then less discoverable but do we need them to be prominent in the docs anyway?
Screenshot 2024-03-18 at 10 34 57 AM copy

### 2. Make navbar one panel (most work)
Currently, the top part of the website is divided into two panels, which steals a bit of the real estate. If we make it one panel, we could use the space more efficiently, for example:
Screenshot 2024-03-18 at 10 34 57 AM

### 3. Add search bar to the left panel of the nav bar
I'm not sure how intuitive and visually pleasing it is but adding it here for the sake of exploration:
Screenshot 2024-03-18 at 10 46 07 AM

### 4. Remove one of the "sign up/sign in" buttons
In doing so, we would save some space but add one additional screen for some of the users -- for example, if we keep "Sign in", the folks without account will have to then click "don't have an account yet? sign up", which is a pattern on some websites.
Screenshot 2024-03-18 at 10 51 48 AM

### 5. Add an additional menu
We could also add a secondary menu for the tabs but this could perhaps clash with the breadcrumbs we will introduce later?

Screenshot 2024-03-18 at 12 13 49 PM

## Solutions -- mobile
Perhaps the four tabs could be a constant element at the top of the menu followed by the sections of the current tab?
Screenshot 2024-03-18 at 10 58 46 AM

## Other Questions
I am not yet sure if "Getting started" should be a tab.
- **What problem we are solving here:** 
  - with different tabs, this important section (containing quickstarts, explanation what Inngest is, how it works, etc.) may be less discoverable if it's a section within "Guides" menu. 
  - Not everyone will land on our docs through the homepage or Guides so by placing it as a tab, we ensure everyone who needs it will see it
-  **Cons:** if this is a tab, it takes space in the nav bar
I worry it is taking too much space or attention but at the same time, we need people to quickly find all the information they need. Opinions?
</detail>

### DIFF
--- a/shared/Docs/Navigation.tsx
+++ b/shared/Docs/Navigation.tsx
@@ -48,7 +48,7 @@ export function TabItem({ href, children, matcher }) {
   const isActive = matcher.test(pathname);
   return (
     <li>
-      <a
+      <Link
         href={href}
         className={clsx(
           "text-sm leading-5 transition whitespace-nowrap px-3 py-4 relative top-0.5",
@@ -59,7 +59,7 @@ export function TabItem({ href, children, matcher }) {
         )}
       >
         <span className="relative -top-0.5">{children}</span>
-      </a>
+      </Link>
     </li>
   );
 }
--- a/shared/Docs/Navigation.tsx
+++ b/shared/Docs/Navigation.tsx
@@ -1,4 +1,4 @@
-import { useMemo, useRef } from "react";
+import { useEffect, useMemo, useRef, useState } from "react";
 import Link from "next/link";
 import { useRouter } from "next/router";
 import clsx from "clsx";
@@ -51,9 +51,9 @@ export function TabItem({ href, children, matcher }) {
       <Link
         href={href}
         className={clsx(
-          "text-sm leading-5 transition whitespace-nowrap px-3 py-4 relative top-0.5",
+          "font-medium text-sm leading-5 transition whitespace-nowrap px-3 py-4 relative top-0.5",
           isActive &&
-            "font-medium text-indigo-700 dark:text-white border-b dark:border-b-white border-b-indigo-700  hover:text-indigo-900",
+            "text-indigo-700 dark:text-white border-b dark:border-b-white border-b-indigo-700  hover:text-indigo-900",
           !isActive &&
             "text-slate-600  dark:text-slate-400 hover:text-slate-900 dark:hover:text-white"
         )}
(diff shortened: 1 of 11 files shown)


shared/Docs/navigationStructure.ts
```diff
@@ -485,6 +546,323 @@ const sectionPythonReference: NavGroup[] = [
   },
 ];
 
+const sectionReference: NavGroup[] = [
+  {
+    title: "TypeScript SKD",
```
djfarrelly, line 551:
```suggestion
    title: "TypeScript SDK",
```

shared/Docs/Navigation.tsx
```diff
@@ -20,14 +21,44 @@ function useInitialValue(value, condition = true) {
...
   return (
-    <li className="lg:hidden">
+    <li>
       <Link
         href={href}
-        className="block py-1 text-sm text-slate-600 transition hover:text-slate-900 dark:text-slate-400 dark:hover:text-white"
+        className={clsx(
+          "text-sm leading-5 transition whitespace-nowrap px-3 py-4 relative top-0.5",
+          isActive &&
+            "font-medium text-indigo-700 dark:text-white border-b dark:border-b-white border-b-indigo-700  hover:text-indigo-900",
```
djfarrelly, line 56: Changing to `font-medium` based on the active state causes the tabs to get wider, I think we should make these always `font-medium`

djfarrelly (approved):
Approving, but I still suggest that we change the nav links to always font-medium to avoid the subtle page shift, but it's not blocking!

shared/Docs/Navigation.tsx
sylwiavargas: Thank you!

PR #730, opened 2024-03-26
Page-specific menu: highlights and nested items

sylwiavargas:
# Overview

This PR introduces section highlighting in the page-specific side menu, as well as nested list items (h1, h2, h3).

➡️ Preview

## Considerations

The previous implementation (theme's default) did not work because:
- The list items were truncated
  - this is not great for discoverability or accessibility, and would be a terrible experience on Reference pages where many items start with the same name (for example, `step.`)
- The list was not nested
  - We need the list to be nested because of how Reference pages will be organized

Moreover:
- The page needs to be responsive (if we resize the window, the side menu should recalculate)

## Code quality

There's a weird bit where I use `useState` in `useEffect` which doesn't feel very elegant. This is done to ensure that the correct hight and position of the highlight is calculated. I'm out of ideas here and I welcome corrections.

## Screenshots

The screenshots show four scenarios:
- light/dark mode
- nestedness
- one/many elements visible at the same time
- window resizing / menu relalculates

### Light mode
Screenshot 2024-03-26 at 9 02 23 PM

### Dark mode
Screenshot 2024-03-26 at 9 02 35 PM

### Reference page
An example why a three-level menu is needed:
Screenshot 2024-03-26 at 9 04 14 PM

### Window resizing
https://github.com/inngest/website/assets/45401242/de6b07aa-a345-4290-a152-909d53774b1b

### DIFF
--- a/shared/Docs/Navigation.tsx
+++ b/shared/Docs/Navigation.tsx
@@ -93,7 +93,7 @@ function NavLink({
       target={target}
       className={clsx(
         "flex justify-between items-center gap-2 py-1 pr-3 text-sm font-medium transition group", // group for nested hovers
-        isTopLevel ? "pl-0" : isAnchorLink ? "pl-3" : "pl-4",
+        isTopLevel || isAnchorLink ? "pl-0" : "pl-4",
         active
           ? "text-slate-900 dark:text-white"
           : "text-slate-600 hover:text-slate-900 dark:text-slate-400 dark:hover:text-white",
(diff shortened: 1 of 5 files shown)


shared/Docs/Navigation.tsx
```diff
@@ -175,43 +171,72 @@ function ActivePageMarker({ group, pathname }) {
...
+    };
+  }, []);
+
+  useEffect(() => {
+    if (pageSectionsEl) {
+      setPageSectionListItems([
+        ...(pageSectionsEl?.querySelectorAll("li") ?? []),
+      ]);
+    }
+  }, [router.pathname, windowWidth, pageSectionsEl]);
```
sylwiavargas, line 195: this is the "weird bit" -- see the "Code Quality" section in the [PR description](https://github.com/inngest/website/pull/730)

djfarrelly (approved):
I'll preemptively approve this PR, but I think we need to fix the inconsistent sections that toggle open closed as you move between tabs as we saw during our call earlier today. Pre-approving if you can fix that tomorrow we can merge!

PR #760, opened 2024-04-25
Docs: How Inngest works - Durable Execution

djfarrelly:
A new doc page to learn how Inngest works, and how it works with durable execution.

Specifically, this section also should help as the single linkable definition for blog posts and our website content. This also is ideal to have a clear section for search optimization purposes.

### DIFF
--- a/pages/docs/learn/durable-execution.mdx
+++ b/pages/docs/learn/durable-execution.mdx
@@ -4,7 +4,7 @@ export const description = 'Learn how Durable Execution works with Inngest. '
 
 One of the core features of Inngest is Durable Execution. Durable Execution allows your functions to be fault-tolerant and resilient to failures. The end result is that your code, and therefore, your overall application, is more reliable.
 
-In this guide, we'll cover the basics of Durable Execution, how it works, and how it works with Inngest functions.
+In this guide, we'll cover what Durable Execution is, how it works, and how it works with Inngest functions.
 
 {/* Note - this page is written a specific way for search optimization */}
 ## What is Durable Execution?
@@ -23,7 +23,7 @@ Some key aspects of Durable Execution include:
 Inngest functions are durable as described in the previous section. Inngest functions use "Steps" to define the execution flow of a function. Each step:
 
 * Is a unit of work that can be run and retried independently.
-* Captures any error or exception thrown within it to
+* Captures any error or exception thrown within it to.
 * Returns state (_data_) that can be used by subsequent steps.
 * Can be executed in parallel or sequentially, depending on the function's configuration.
 
--- a/pages/docs/learn/durable-execution.mdx
+++ b/pages/docs/learn/durable-execution.mdx
@@ -43,8 +43,57 @@ Inngest functions are executed incrementally, _step by step_. As functions are e
 Let's look at an example of a function and walk through how it is executed:
 
 ```typescript
-await step.run("create-user", async () => {
-  const user = await createUser(state);
-  return { user };
-});
+const fn = inngest.createFunction(
+  { id: "import-contacts" },
+  { event: "contacts/csv.uploaded" },
+  // The function handler:
+  async ({ event, step }) => {
+    const rows = await step.run("parse-csv", async () => {
+      return await parseCsv(event.data.fileURI);
+    });
+
+    const normalizedRows = await step.run("normalize-raw-csv", async () => {
+      const normalizedColumnMapping = getNormalizedColumnNames();
+      return normalizeRows(rows, normalizedColumnMapping);
+    });
+
+    const results = await step.run("input-contacts", async () => {
+      return await importContacts(normalizedRows);
+    });
+
+    return { results };
+  }
 ```
+
+### Initial execution
+
+1. When the function is first called, the _function handler_ is called with only the `event` payload data sent.
+2. When the first step is discovered, the `"parse-csv"` step is run. As the step has not been executed before, the step's code (the callback function) is run and the result is captured.
+3. The function does not continue executing beyond this step. Each SDK uses a different method to interrupt the function execution before running any more code in your function handler.
+4. Internally, the step's id (`"parse-csv"`) is hashed as the state identifier to be used in future executions. Additionally, the steps' index (`0` in this case) is also included in the result.
+5. The result is sent back to Inngest and persisted in the function state store.
+
+### Secondary executions - Memoization of steps
+
+Steps executed after the first step leverage the state of previous executions and memoization. Here's how it works:
+
+6. The function is re-executed, this time with the `event` payload data and the state of the previous execution in JSON.
+7. The next step is discovered (`"parse-csv"`).
+8. The previous result is found in the state of previous executions. Internally, the SDK uses the hash of the step name to look up the result in the state data.
+9. The step's code is not executed, instead the SDK injects the result into the return value of `step.run`, (in this example, the data will be returned as `rows`).
+10. The function continues execution until the next step is discovered (`"normalize-raw-csv"`).
+11. The step's code is executed and the result is returned to Inngest (in the same approach as steps 2-5 above).
+
+### Error handling
+
+Some steps may throw errors or exceptions during execution. Here's how error handling works within function execution:
+
+12. If an error occurs during the execution of a step (for example, `"input-contacts"`), the function is interrupted and the error is caught by the SDK.
+13. The error is serialized and returned to Inngest. The number of attempts are logged and the error is persisted in the function state store.
+14. Depending on the number of attempts configured for the function, the function may be retried (see: [Error handling]((/docs/guides/error-handling))):
+    * If the the function _has not_ exhausted the number of attempts, the function is re-executed from the point of failure with the state of all previous step executions. The step is re-executed and follows the same process as above (see: steps 6-11).
+    * If the function _has_ exhausted the number of attempts, the function is re-executed with the error thrown. The function can then catch and handle the error as desired (see: [Handling a failing step](https://www.inngest.com/docs/guides/error-handling#handling-a-failing-step)).
+
+
+{/* TODO - Add what happens when steps are re-ordered (or added) */}
+{/* TODO - Add what happens when a step is removed */}
--- a/pages/docs/learn/durable-execution.mdx
+++ b/pages/docs/learn/durable-execution.mdx
@@ -9,26 +9,42 @@ In this guide, we'll cover what Durable Execution is, how it works, and how it w
 {/* Note - this page is written a specific way for search optimization */}
 ## What is Durable Execution?
 
-Durable Execution is a fault-tolerant approach to executing code that handles failures and interruptions gracefully through automatic retries and state persistence. This means that your code can continue to run even if there are issues like network failures, timeouts, infrastructure outages, and other transient errors.
+Durable Execution is a fault-tolerant approach to executing code that is achieved by handling failures and interruptions gracefully with automatic retries and state persistence. This means that your code can continue to run even if there are issues like network failures, timeouts, infrastructure outages, and other transient errors.
 
-Durable functions can throw errors or exceptions and automatically retry, resuming execution from the point of failure. Durable functions are designed to be long-running and stateful, meaning that they can persist state across function invocations and retries.
-
-Some key aspects of Durable Execution include:
+Key aspects of Durable Execution include:
 
 * **State persistance** - Function state is persisted outside of the function execution context. This enables function execution to be resumed from the point of failure on the same _or_ different infrastructure.
 * **Fault-tolerance** - Errors or exceptions are caught by the execution layer and are automatically retried. Retry behavior can be customized to handle the accepted number of retries as well as handle different types of errors.
 
+In practice, Durable Execution is implemented in the form of "durable functions," sometimes also called "durable workflows." Durable functions can throw errors or exceptions and automatically retry, resuming execution from the point of failure. Durable functions are designed to be long-running and stateful, meaning that they can persist state across function invocations and retries.
+
 ## How Inngest functions work
 
-Inngest functions are durable as described in the previous section. Inngest functions use "Steps" to define the execution flow of a function. Each step:
+Inngest functions are durable: they throw errors or exceptions, automatically retry from the point of failure, and can be stateful and long-running.
+
+Inngest functions use "**Steps**" to define the execution flow of a function. Each step:
 
 * Is a unit of work that can be run and retried independently.
 * Captures any error or exception thrown within it to.
+* Will not be re-executed if it has already been successfully executed.
 * Returns state (_data_) that can be used by subsequent steps.
 * Can be executed in parallel or sequentially, depending on the function's configuration.
 
-In comparison, some durable execution systems modify the runtime environment to support durable execution. Inngest functions are designed to be stateless and can be run in any environment that supports the Inngest runtime.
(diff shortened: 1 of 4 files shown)


pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,34 @@
+export const description = 'Learn how Durable Execution works with Inngest. '
+
+# How Durable Execution works with Inngest
+
+One of the core features of Inngest is Durable Execution. Durable Execution allows your functions to be fault-tolerant and resilient to failures. The end result is that your code, and therefore, your overall application, is more reliable.
+
+In this guide, we'll cover the basics of Durable Execution, how it works, and how it works with Inngest functions.
```
sylwiavargas, line 7:
```suggestion
In this guide, we'll cover what Durable Execution is, how it works, and how it works with Inngest functions.
```

this would be less ego-triggering for some folks

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+{/* Note - this page is written a specific way for search optimization */}
+## What is Durable Execution?
+
+Durable Execution is a fault-tolerant approach to executing code that handles failures and interruptions gracefully through automatic retries and state persistence. This means that your code can continue to run even if there are issues like network failures, timeouts, infrastructure outages, and other transient errors.
```
sylwiavargas, line 12:
```suggestion
Durable Execution is a fault-tolerant approach to executing code. Durable Execution handles failures and interruptions gracefully through automatic retries and state persistence. This means that your code can continue to run even if there are issues like network failures, timeouts, infrastructure outages, and other transient errors.
```

sylwiavargas, line 12: I'd split the first sentence into two here. Short sentences are preferable when explaining new concepts. (thinking of an alternative that also doesn't involve redundant duplication or a passive voice 🤪 )

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+Durable functions can throw errors or exceptions and automatically retry, resuming execution from the point of failure. Durable functions are designed to be long-running and stateful, meaning that they can persist state across function invocations and retries.
```
sylwiavargas, line 14: "Durable functions" are a new term. Should we add something like:

```suggestion
At the foundation of Durable Execution are the so-called "durable functions". Durable functions can throw errors or exceptions and automatically retry, resuming execution from the point of failure. They are designed to be long-running and stateful, meaning that they can persist state across function invocations and retries.
```

sylwiavargas: Additionally, it feels like this paragraph belongs at the end of this section (after "Some key aspects" which expands on the previous paragraph).

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+Some key aspects of Durable Execution include:
+
+* **State persistance** - Function state is persisted outside of the function execution context. This enables function execution to be resumed from the point of failure on the same _or_ different infrastructure.
+* **Fault-tolerance** - Errors or exceptions are caught by the execution layer and are automatically retried. Retry behavior can be customized to handle the accepted number of retries as well as handle different types of errors.
+
+## How Inngest functions work
+
+Inngest functions are durable as described in the previous section. Inngest functions use "Steps" to define the execution flow of a function. Each step:
```
sylwiavargas, line 23:
```suggestion
Inngest functions are durable: they throw errors or exceptions, automatically retry from the point of failure, and can be stateful and long-running.

Inngest functions use "Steps" to define the execution flow of a function. Each step:
```

SEO-friendly + more linkable

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+* Is a unit of work that can be run and retried independently.
+* Captures any error or exception thrown within it to
```
sylwiavargas, line 26:
```suggestion
* Captures any error or exception thrown within it to.
```

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,34 @@
...
+* Returns state (_data_) that can be used by subsequent steps.
+* Can be executed in parallel or sequentially, depending on the function's configuration.
+
+In comparison, some durable execution systems modify the runtime environment to support durable execution. Inngest functions are designed to be stateless and can be run in any environment that supports the Inngest runtime.
```
sylwiavargas, line 30: This is not clear to me: in the previous section, we are saying that durable functions are designed to be stateful but this paragraph states the opposite.

sylwiavargas: "any environment that supports the Inngest runtime" -- I'd add the most common ones as examples just so this doesn't sound like Inngest is a tool that requires niche config.

sylwiavargas: btw. should we continue to capitalize "Durable Execution"?

pages/docs/learn/durable-execution.mdx
djfarrelly: Yeah - I reworked this a lot, but I was thinking that the approach (the how) is important to what it is. Instead of "through" it could be "achieved with" which may be easier to read 🤔 /

pages/docs/learn/durable-execution.mdx
djfarrelly: You're right - I think this was a placeholder / work in progress. I'll change it!

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,99 @@
+export const description = 'Learn how Durable Execution works with Inngest. '
+
+# How Durable Execution works with Inngest
+
+One of the core features of Inngest is Durable Execution. Durable Execution allows your functions to be fault-tolerant and resilient to failures. The end result is that your code, and therefore, your overall application, is more reliable.
+
+In this guide, we'll cover what Durable Execution is, how it works, and how it works with Inngest functions.
```
sylwiavargas, line 7:
```suggestion
This page covers what Durable Execution is, how it works, and how it works with Inngest functions.
```

(not a guide)

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+{/* Note - this page is written a specific way for search optimization */}
+## What is Durable Execution?
+
+Durable Execution is a fault-tolerant approach to executing code that is achieved by handling failures and interruptions gracefully with automatic retries and state persistence. This means that your code can continue to run even if there are issues like network failures, timeouts, infrastructure outages, and other transient errors.
+
+Key aspects of Durable Execution include:
+
+* **State persistance** - Function state is persisted outside of the function execution context. This enables function execution to be resumed from the point of failure on the same _or_ different infrastructure.
+* **Fault-tolerance** - Errors or exceptions are caught by the execution layer and are automatically retried. Retry behavior can be customized to handle the accepted number of retries as well as handle different types of errors.
```
sylwiavargas, line 17:
```suggestion
* **Fault-tolerance** - Errors or exceptions are caught by the execution layer and are automatically retried. Retry behavior can be customized to handle the accepted number of retries and different types of errors.
```

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+Inngest functions use "**Steps**" to define the execution flow of a function. Each step:
+
+* Is a unit of work that can be run and retried independently.
+* Captures any error or exception thrown within it to.
+* Will not be re-executed if it has already been successfully executed.
+* Returns state (_data_) that can be used by subsequent steps.
+* Can be executed in parallel or sequentially, depending on the function's configuration.
+
+Complex functions can be comprised of many steps. This allows a long-running function to be broken down into smaller, more manageable units of work. As each step is retried independently, and the function can be resumed from the point of failure, avoiding unnecessary re-execution of work.
```
sylwiavargas:
```suggestion
Complex functions can consist of many steps. This allows a long-running function to be broken down into smaller, more manageable units of work. As each step is retried independently, and the function can be resumed from the point of failure, avoiding unnecessary re-execution of work.
```

sylwiavargas: plain English

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+In comparison, some durable execution systems modify the runtime environment to persist state or interrupt errors or exceptions. Inngest functions are defined with steps written in normal language code, which enables them to run in any environment or runtime - including serverless environments - without modification.
```
sylwiavargas:
```suggestion
In comparison, some Durable Execution systems modify the runtime environment to persist state or interrupt errors or exceptions. Inngest functions are defined with steps written in standard language code, which enables them to run in any environment or runtime - including serverless environments - without modification.
```

or "common"? there's a lot of discussion about referring to the category of "normalcy" so it's better to just avoid it unless it about normal/abnormal issues.

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+### How steps are executed
+
+Inngest functions are defined with a series of steps that define the execution flow of the function. Each step is defined with a unique ID and a function that defines the work to be done. The data returned can be used by subsequent steps.
+
+Inngest functions are executed incrementally, _step by step_. As functions are executed, the results of each step are returned to Inngest and persisted in a managed function state store. The function is then re-executed and any steps that have already been successfully executed are [_memoized_](https://en.wikipedia.org/wiki/Memoization). This means that the successfully executed steps are skipped and the SDK injects the data returned by the previous execution into the function.
```
sylwiavargas: not a blocker but these two paragraphs could really use a flowchart

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+Let's look at an example of a function and walk through how it is executed:
```
sylwiavargas: not a blocker but a note to self: this section could use a tool like CodeHike to explain this code as a process

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,99 @@
...
+    return { results };
+  }
+```
+
+### Initial execution
+
+1. When the function is first called, the _function handler_ is called with only the `event` payload data sent.
+2. When the first step is discovered, the `"parse-csv"` step is run. As the step has not been executed before, the step's code (the callback function) is run and the result is captured.
+3. The function does not continue executing beyond this step. Each SDK uses a different method to interrupt the function execution before running any more code in your function handler.
+4. Internally, the step's id (`"parse-csv"`) is hashed as the state identifier to be used in future executions. Additionally, the steps' index (`0` in this case) is also included in the result.
```
sylwiavargas:
```suggestion
4. Internally, the step's ID (`"parse-csv"`) is hashed as the state identifier to be used in future executions. Additionally, the steps' index (`0` in this case) is also included in the result.
```

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,99 @@
...
+      return normalizeRows(rows, normalizedColumnMapping);
+    });
+
+    const results = await step.run("input-contacts", async () => {
+      return await importContacts(normalizedRows);
+    });
+
+    return { results };
+  }
+```
```
sylwiavargas:
```suggestion
    }
  )
```
```

File: pages/docs/learn/durable-execution.mdx [hunk 15]
```diff
@@ -0,0 +1,99 @@
...
+      const normalizedColumnMapping = getNormalizedColumnNames();
+      return normalizeRows(rows, normalizedColumnMapping);
+    });
+
+    const results = await step.run("input-contacts", async () => {
+      return await importContacts(normalizedRows);
+    });
+
+    return { results };
+  }
```
sylwiavargas: There's a `)` missing (I can't add a suggestion)

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+1. When the function is first called, the _function handler_ is called with only the `event` payload data sent.
+2. When the first step is discovered, the `"parse-csv"` step is run. As the step has not been executed before, the step's code (the callback function) is run and the result is captured.
+3. The function does not continue executing beyond this step. Each SDK uses a different method to interrupt the function execution before running any more code in your function handler.
+4. Internally, the step's id (`"parse-csv"`) is hashed as the state identifier to be used in future executions. Additionally, the steps' index (`0` in this case) is also included in the result.
+5. The result is sent back to Inngest and persisted in the function state store.
+
+### Secondary executions - Memoization of steps
+
+Steps executed after the first step leverage the state of previous executions and memoization. Here's how it works:
```
sylwiavargas:
```suggestion
Each of the subsequent steps leverages the state of previous executions and memoization. Here's how it works:
```

I don't know if this is a good rephrasing. The original sentence sounds a bit confusing.

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,99 @@
...
+10. The function continues execution until the next step is discovered (`"normalize-raw-csv"`).
+11. The step's code is executed and the result is returned to Inngest (in the same approach as steps 2-5 above).
+
+### Error handling
+
+Some steps may throw errors or exceptions during execution. Here's how error handling works within function execution:
+
+12. If an error occurs during the execution of a step (for example, `"input-contacts"`), the function is interrupted and the error is caught by the SDK.
+13. The error is serialized and returned to Inngest. The number of attempts are logged and the error is persisted in the function state store.
+14. Depending on the number of attempts configured for the function, the function may be retried (see: [Error handling]((/docs/guides/error-handling))):
```
sylwiavargas:
```suggestion
14. Depending on the number of attempts configured for the function, the function may be retried (see: [Error handling](/docs/guides/error-handling)):
```

pages/docs/learn/durable-execution.mdx
```diff
@@ -0,0 +1,99 @@
...
+    * If the the function _has not_ exhausted the number of attempts, the function is re-executed from the point of failure with the state of all previous step executions. The step is re-executed and follows the same process as above (see: steps 6-11).
+    * If the function _has_ exhausted the number of attempts, the function is re-executed with the error thrown. The function can then catch and handle the error as desired (see: [Handling a failing step](https://www.inngest.com/docs/guides/error-handling#handling-a-failing-step)).
```
sylwiavargas:
```suggestion
    * If the function _has_ exhausted the number of attempts, the function is re-executed with the error thrown. The function can then catch and handle the error as desired (see: [Handling a failing step](/docs/guides/error-handling#handling-a-failing-step)).
```

sylwiavargas (approved):
Thank you for writing this page!

I left a few comments here and there.

pages/docs/learn/durable-execution.mdx
djfarrelly: Yeah - visuals would be great for this page, animations too, but we'll at least start with text 😅

pages/docs/learn/durable-execution.mdx
djfarrelly: Re-wrote for clarity - my original sentence didn't convey the real purpose :) good catch

PR #762, opened 2024-04-26
update fly.io's limits

darwin67:
most are not documented so just mark as `???` for now.
we probably should add service like Render and such as well


djfarrelly (changes requested):
In our docs, it's ideal to use exact language whenever possible. Can we change the ambiguous "???" to a more clear "Undocumented"?

djfarrelly (approved):

PR #767, opened 2024-05-02
adds Banger blog post

sylwiavargas:


pages/blog/_posts/banger.mdx
```diff
@@ -0,0 +1,279 @@
...
+Hi everyone! My name is [Igor Samokhovets](https://twitter.com/IgorSamokhovets) and I'm a music producer who goes by the artist name Tequila Funk. Together with Mark Beziaev we created [banger.show](https://banger.show/) -- a video maker app for musicians, DJs, and labels. It allows music industry people to create stunning visual assets for their music.
+
+<YouTube id="eSLKm6IfZM4" height={300}/>
+
+ Creating a video for your new song takes only a few minutes and you don't need to know any complex software because banger.show works in your browser.
+
+
+## Making background processing snappy
+
+At banger.show, we do a lot of background processing. Managing render states, generating visual assets in the background, and even handling the rendering process on remote distributed workers. We chose Inngest because there's no better way to handle background jobs if you're using Next.js.
```
djfarrelly, line 22: > there's no better way to handle background jobs if you're using Next.js

This is fun, but I was wondering if there are specifics we could include, something like: "because it makes it easy to orchestrate multi-step workflows right in your Next.js codebase."

pages/blog/_posts/banger.mdx
```diff
@@ -0,0 +1,279 @@
...
+
+Let's dive into some parts of the video rendering pipeline.
```
djfarrelly, line 24: It would be nice for a little context, like "this pipeline receives an uploaded mp3 from the user and combines the audio with a generated visualization designed to match the audio to create a unique video for the user's music"

pages/blog/_posts/banger.mdx
```diff
@@ -0,0 +1,279 @@
...
+
+### 1. Updating the render status
+
+The first step in the render pipeline is to update the render status and set user credits on hold
```
djfarrelly, line 28: Should we show the `inngest.createFunction` part of this? Just for context. Otherwise it's kind of dangling. Otherwise, we can say something like - "for this blog post, we're going to focus on the steps of our Inngest function" then link to docs which cover multiple step functions.

pages/blog/_posts/banger.mdx
```diff
@@ -0,0 +1,279 @@
...
+    }
+  )
+
+  return video
+})
+```
+
+### 2. Cropping the audio file
+
+In banger.show, user should select a fragment of the song to create a videoization for it. In the background job, we:
```
djfarrelly, line 60:
```suggestion
In banger.show, the user selects a fragment of the song to create a "videoization" for it. In the background job, we:
```
Added quotes b/c I would guess videoization is their own amalgamation of video and visualization.

pages/blog/_posts/banger.mdx
```diff
@@ -0,0 +1,279 @@
...
+At banger.show, we do a lot of background processing. Managing render states, generating visual assets in the background, and even handling the rendering process on remote distributed workers. We chose Inngest because there's no better way to handle background jobs if you're using Next.js.
+
+Let's dive into some parts of the video rendering pipeline.
+
+### 1. Updating the render status
+
+The first step in the render pipeline is to update the render status and set user credits on hold
+
+```js
+const updatedVideo = await step.run('Update user balance', async () => {
```
djfarrelly, line 31:
```suggestion
const updatedVideo = await step.run('update-user-balance', async () => {
```
It's a best practice to use slugs as these are IDs (maybe a legacy of our old v1/v2 SDK), @sylwiavargas can we update all step ids to slugs?

pages/blog/_posts/banger.mdx
```diff
@@ -0,0 +1,279 @@
...
+    })
+
+    return renderedVideo
+  }
+)
+```
+
+### 4. Invalidate CDN cache
+
+There are some cases when video needs to be re-rendered. To make sure CDN cache is always fresh, we purge it each time the video renders.
```
djfarrelly, line 198: Extra context like "We host our videos on a CDN, but some times, the video needs to be re-rendered. To make sure..."

pages/blog/_posts/banger.mdx
djfarrelly: Another thing we could add is "why use steps" for this - basically saying, that this allows it to run on serverless or sometimes things fail, etc. Are we using concurrency settings for this at all?

djfarrelly (commented):
Some suggestions inline to add more context for the reader. It would also be nice to add more links, to our docs or otherwise!

pages/blog/_posts/banger.mdx
samohovets: Hi Dan! This would be more descriptive and correct:

> We chose Inngest because there's no better way to handle background jobs if you're using Next.js **without a custom server**. **It's a primary "flow controller" for us**, even though we have a simple queue solution based on Redis to handle tasks on our infra. It allows us to orchestrate, observe, and abstract away lower-level processes.

In simple terms Inngest allows us to operate on all of this low-level Redis stuff on a higher level, e.g. 

- not delegate every post-render task to the video rendering machine, but have some flow where the main server can receive render results from the workers and do something with it
- observe each step and its return data in the dashboard
- handle emergency cases when all of our infra goes down and we need to spin up some backup workers on AWS

sylwiavargas: This is great context, thank you, Igor! I'll add these examples as well!

pages/blog/_posts/banger.mdx
samohovets: Yeah, the current version was missing some context:

- We're at the default Inngest concurrency
- We have a graceful flow termination strategy in the `onFailure` function. It's pretty simplified here, since each thing we have to share brings up additional questions on how the whole thing works and the blog post would fill up with boring / verbose stuff.

```ts
export const renderVideo = inngest.createFunction(
  {
    name: 'Render video',
    id: 'render-video',
    cancelOn: [
      {
        event: 'banger/video.create',
        match: 'data.videoId'
      }
    ],
    onFailure: async ({ error, event, step }) => {
      await dbConnect()

      const isStalled = RenderStalledError.isRenderStalledError(error)

       const updatedVideo = await step.run(
        'Update video status to failed',
        () =>
          VideoModel.findOneAndUpdate(
            { _id: event.data.event.data.videoId },
            {
              $set: {
                status: isStalled ? 'stalled' : 'error',
                ...(isStalled && { stalledAt: new Date() }),
                renderProgress: null
              }
            },
            { new: true }
          )
            .lean()
      )

      invariant(updatedVideo, 'no video found')

      // refund user units if error is not recoverable
      // if it's stalled, we're going to recover it later
      if (!isStalled) {
        await step.run('Refund user units', async () => {
          await UserModel.updateOne(
            {
              _id: event.data.event.data.userId
            },
            { $inc: { unitsRemaining: updatedVideo.videoDuration } }
          )
        })
      }

      if (process.env.NODE_ENV === 'production') {
        const errorJson = _.truncate(JSON.stringify(event), {
          length: 3000
        })
        await sendTelegramLog(
          _.truncate(
            `🚨 Error while rendering video: ${error.message}\n
          Event: ${errorJson}\n`,
            { length: 3000 }
          )
        )
      }

      Sentry.captureException(error)
    }
  },
  { event: 'banger/video.create' },
  async ({ event, step, attempt, logger }) => {
    // ...
  })

File: pages/blog/_posts/banger.mdx (hunk 2 above)
Comment (samohovets) [reply pages/blog/_posts/banger.mdx:30]: @sylwiavargas I'm not sure how to put it simply:

- The pipeline receives the project created by the user
- Converts the audio for the more efficient rendering
- Controlling the flow of the render: sending user project to the render machine, updating statuses and progress, handling error retries and "stalled render" cases
- Waiting for the render to finish and perform post-render tasks like invalidating CDN cache, creating video thumbnails, sending email to the user when the video is ready

File: pages/blog/_posts/banger.mdx [hunk 7]
```diff
@@ -3,61 +3,123 @@ heading: A Deep Dive into a Video Rendering Pipeline
...
+      }
 
-  return video
-})
+      Sentry.captureException(error)
+    }
+  },
+  { event: 'banger/video.create' },
+  async ({ event, step, attempt, logger }) => {
+    // ...
```
djfarrelly, line 113: The old code which updated the user status (the name of this section) is now missing, is this correct? Do we mean to bring back that first step?

The `onFailure` handler is cool, but perhaps we should just move that to the end of the post in it's own section and link to our docs on `onFailure`.

PR #778, opened 2024-05-14
Update pages with duplicate names

djfarrelly:
These pages had the same name which caused confusion in docs search and likely SEO weirdness. The guide should be the goto link with the reference only living on purely for reference. We may delete the reference at some point in the future.


sylwiavargas (approved):
Thank you!

PR #783, opened 2024-05-17
Changes the color of the "Previous" / "Next" buttons

sylwiavargas:
The previous color combo doesn't look great, and is not very accessbile

### BEFORE
Screenshot 2024-05-17 at 2 06 43 PM

### AFTER
Screenshot 2024-05-17 at 2 19 01 PM

### DIFF
--- a/shared/Button.tsx
+++ b/shared/Button.tsx
@@ -27,8 +27,11 @@ export function Button({
   };
 
   const variants = {
-    primary: "text-white bg-indigo-500 hover:bg-indigo-400",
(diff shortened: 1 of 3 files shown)


djfarrelly (commented):
It's kind of weird using the primary button on non-CTAs. Perhaps we could improve our `secondary` button color instead?

djfarrelly:
Made a quick variant for secondary better for contrast. It's only used in the docs so it's a safe, contained change :) 

image

djfarrelly (approved):

PR #784, opened 2024-05-17
Add mdx to tailwind config

sylwiavargas:
Previously, only the Tailwind classes used elsewhere were recognized in the docs and blog, now any Tailwind classes can be used in `.mdx` files.


djfarrelly (approved):
Good catch! I probably banged my head against the wall with this more than once!

PR #786, opened 2024-05-21
Example: Track failures in Datadog

djfarrelly:
This better surfaces a popular question that we have in examples.

## Questions for reviewers:

* Should this be more generic, like "Tracking function failures in external tools"?
* Should we adjust the nav sidebar to handle longer page names?

Screen Shot 2024-05-21 at 12 55 51 PM


sylwiavargas:
> Should we adjust the nav sidebar to handle longer page names?

Oh, definitely. It could just wrap the names.
Do you want to include it in the PR/send a new one? If not, I can do it tomorrow.

sylwiavargas (commented):
Thank you for adding this example 🙏

PR #791, opened 2024-05-24
move "Security" to "/learn/security"

sylwiavargas:
This seems a better home for this page as it does pertain to how Inngest works.

Screenshot 2024-05-24 at 2 01 19 PM


sylwiavargas:
@djfarrelly how do you feel about this change?

next.config.mjs
```diff
@@ -36,6 +36,9 @@ const permanentRedirects = [
   ["/uses/zero-infra-llm-ai", "/ai"],
   ["/uses/internal-tools", "/uses/workflow-engine"],
   ["/uses/user-journey-automation", "/blog/lifecycle-emails-with-resend"],
+
+  // new IA
+  ["/docs/security", "docs/learn/security"],
```
djfarrelly, line 41:
```suggestion
  ["/docs/security", "/docs/learn/security"],
```

djfarrelly (approved):
I think this works - feels like it's related to deeper system investigation rather than learning the basics, so happy for this to be relocated. 

Left one required change prior to merging to fix the redirect

next.config.mjs
sylwiavargas: Ohhhh good catch, thank you!

PR #801, opened 2024-06-05
Docs: remove v2 snippets from guides

sylwiavargas:
This PR removes the `v2` tab from the code snippets in our guides.

I have not deleted it in the reference because I assumed that if someone needs the information on v2/v3, reference section would be the correct place to look for it. That being said, it's been 8 months so maybe it's time to bid it adieu altogether? Wdyt @djfarrelly?


djfarrelly (approved):
> I have not deleted it in the reference because I assumed that if someone needs the information on v2/v3, reference section would be the correct place to look for it. That being said, it's been 8 months so maybe it's time to bid it adieu altogether? Wdyt @djfarrelly?

Yep @sylwiavargas - let's keep it in the reference and upgrade guides for now. The cleanup in this PR to remove from guides makes them a lot cleaner and will pave the way for us potentially adding Python code examples in select guides as well :)

PR #832, opened 2024-07-03
Add steps to the example

sylwiavargas:
Adds steps into the code example:
Screenshot 2024-07-03 at 11 30 54 AM

### DIFF
--- a/pages/docs/learn/inngest-functions.mdx
+++ b/pages/docs/learn/inngest-functions.mdx
@@ -26,10 +26,13 @@ export default inngest.createFunction(
   async ({ event, step }) => {
     // Here goes the business logic
     // By wrapping code in steps, it will be retried automatically on failure
-    const copyImages = await step.run("copy-images-to-s3", async () => {
+    const s3Urls = await step.run("copy-images-to-s3", async () => {
       return copyAllImagesToS3(event.data.imageURLs);
     });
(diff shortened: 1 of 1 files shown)


pages/docs/learn/inngest-functions.mdx
```diff
@@ -24,10 +24,12 @@ export default inngest.createFunction(
...
-    // here goes the business logic
-    return copyAllImagesToS3(event.data.imageURLs);
-    // if you need to perform more than one task,
-    // wrap each one in step.run()
+    // Here goes the business logic
+    // By wrapping code in steps, it will be retried automatically on failure
+    const copyImages = await step.run("copy-images-to-s3", async () => {
+      return copyAllImagesToS3(event.data.imageURLs);
+    });
+    // You can include numerous steps in your function
```
djfarrelly, line 32: It might be good to show something using the result of the first step. This might be easier to illustrate the concept of oh this is what goes in a step. Without it, a single step feels a little unclear I think

```suggestion
    const s3Urls = await step.run("copy-images-to-s3", async () => {
      return copyAllImagesToS3(event.data.imageURLs);
    });
    // You can include numerous steps in your function
    await step.run('resize-images', async () => {
      await resizer.bulk({ urls: s3Urls, quality: 0.9, maxWidth: 1024 });
    })
```

PR #833, opened 2024-07-03
fix(docs): minor broken links and typos

charlypoly:
Minor stuff I've encountered while going through the docs that might be worth fixing before the IA project lands 👀


sylwiavargas (approved):
Fantastic, thank you, detective! 🙏

PR #834, opened 2024-07-03
Follow #831 to add noindex tag

djfarrelly:
Follows #831 off the advice of [this blog post](https://searchengineland.com/ppc-landing-pages-seo-388033) to properly prevent indexing of these pages but allow GTM.


sylwiavargas (approved):
TIL, thank you

PR #842, opened 2024-07-05
Remove slugify call for app IDs

goodoldneon:
We don't want users calling `slugify` on app IDs since it could change the app ID, which will make us think it's a new app


pages/docs/sdk/migration.mdx
```diff
@@ -168,7 +168,7 @@ When instantiating a client using `new Inngest()` or creating a function via `in
           import { Inngest, slugify } from "inngest";
```
sylwiavargas, line 168:
```suggestion
          import { Inngest } from "inngest";
```

should we also remove `slugify` here?

sylwiavargas (approved):
Approving but left a comment 💜

PR #864, opened 2024-07-19
fix(docs): ensure Python SDK deeply nested menu items stay visible when active

charlypoly:
## Before

_The Python SDK section closes when navigating to the nested Steps items_

image

## After

_The Python SDK section remains open when navigating to the nested Steps items_

image


sylwiavargas (approved):
Thank you!

I did a [WIP fix](https://github.com/inngest/website/pull/775) some time ago (between conferences so it stayed WIP forever). That had some tradeoffs and it didn't occur to me we could approach it this way 😂

PR #865, opened 2024-07-22
Update the RetryAfterError documentation

ValentinH:
While testing the RetryAfterError, I discovered that the actual behaviour wasn't matching the docs.
Based on the implementation: https://github.com/inngest/inngest-js/blob/main/packages/inngest/src/components/RetryAfterError.ts

If the `retryAfter` value is a number, it should be in milliseconds.


sylwiavargas (approved):
Good catch! We recently updated the comments you refer to in the package but clearly, forgot to do that here. Thank you 🙏

PR #868, opened 2024-07-24
Add Python encryption docs

goodoldneon:
- Add Python encryption docs
- Fix broken env var defaults in `package.json`


jpwilliams (approved):

pages/docs/reference/python/middleware/encryption.mdx
```diff
@@ -0,0 +1,34 @@
...
+```
+
+The following data is encrypted by default:
+- The `encrypted` field in `event.data`.
+- `step.run` return values.
+- Function return values.
+
+## Changing the encrypted `event.data` field
+
+By default, `event.data.encrypted` is encrypted. All other fields are sent in plaintext. If you'd like to encrypt a different field, set the `event_encryption_field` parameter.
```
sylwiavargas, line 22:
```suggestion
By default, `event.data.encrypted` is encrypted. All other fields are sent in plaintext. To encrypt a different field, set the `event_encryption_field` parameter.
```

- more concise
- here we don't need to "cushion" the message because it's not a distressing news and we are not making a request to a user

pages/docs/reference/python/middleware/encryption.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+## Decrypt only mode
+
+If you'd like to disable encryption but continue decrypting, set `decrypt_only=True`. This is useful when you want to migrate away from encryption but continue processing older events.
```
sylwiavargas, line 26:
```suggestion
To disable encryption but continue decrypting, set `decrypt_only=True`. This is useful when you want to migrate away from encryption but still need to process older events.
```

- same as above
- nit: easier mental parsing thanks to "want to"/"need to" mirror structure

pages/docs/reference/python/middleware/encryption.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+## Fallback decryption keys
+
+If you'd like to attempt decryption with multiple keys, set the `fallback_decryption_keys` parameter. This is useful when rotating keys, since older events may have been encrypted with a different key.
```
sylwiavargas, line 30:
```suggestion
To attempt decryption with multiple keys, set the `fallback_decryption_keys` parameter. This is useful when rotating keys, since older events may have been encrypted with a different key.
```

sylwiavargas (approved):
Thank you 🙏
Left a few suggestions, see if they work for you.
Approving already.

PR #878, opened 2024-08-12
Add run state size to limits page

goodoldneon:


pages/docs/usage-limits/inngest.mdx
```diff
@@ -27,6 +27,14 @@ Our free plan has a concurrency limit of `25`, and paid plans have concurrency l
 
 The limit for data returned by a step is `4MB`.
 
+### Function run state size
+
+Function run state cannot exceed `32MB`. Its state includes:
+- Event data (multiple events if using batching)
+- Step-returned data
+- Function-returned data
+- Internal metadata (rarely a noticeable size)
```
charlypoly, line 36: @goodoldneon, do we need to list internal metadata? It doesn't seem like actionable information for the developer. If we keep it, can we provide an estimation of its size?

charlypoly, line 36:
```suggestion
- Internal metadata (_small - around a few bytes_)
```

charlypoly (approved):

PR #891, opened 2024-08-26
docs: fix user defined workflow example

FedericoBiccheddu:


charlypoly (approved):
Thank you for spotting this @FedericoBiccheddu! ⚡

PR #893, opened 2024-08-29
Replace icons

anafilipadealmeida:
Replaces a bunch of icons by remix ones, fixes a bunch of SVG ones
Deleted the folder "docs/icons" and removed most icons inside "shared/Icons". Whenever possible, we should be importing the icons directly from remix.


charlypoly (approved):
Thank you! you rock! ⚡

PR #906, opened 2024-09-12
Add a Pydantic guide

goodoldneon:
Add a Pydantic guide to the Python SDK's reference.

### DIFF
--- a/pages/docs/reference/python/guides/pydantic.mdx
+++ b/pages/docs/reference/python/guides/pydantic.mdx
@@ -1,0 +1,66 @@
+# Pydantic
+
+This guide will help you use Pydantic to perform runtime type validation when sending and receiving events.
+
+## Sending events
+
+Create a base class that all your event classes will inherit from. This class has methods to convert to and from `inngest.Event` objects.
+```py
+import inngest
+import pydantic
+import typing
+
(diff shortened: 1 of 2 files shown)


charlypoly (approved):
Thank you @goodoldneon!

pages/docs/reference/python/guides/pydantic.mdx
```diff
@@ -0,0 +1,66 @@
...
+    name: str = "my-event"
+```
+
+Since Pydantic validates on instantiation, the following code will raise an error if the data is invalid.
+```py
+client.send(
+    MyEvent(
+        data=MyEventData(count="bad data"),
+    ).to_event()
+)
```
djfarrelly, line 50: Can we use a real life event name? We've gotten feedback in the past that real event names and function names are a lot better to learn than "hello world" type stuff. What about:

```py
class PostUpvotedEventData(pydantic.BaseModel):
    count: int

class PostUpvotedEvent(BaseEvent):
    data: MyEventData
    name: str = "forum/post.upvoted"
```

Since Pydantic validates on instantiation, the following code will raise an error if the data is invalid.
```py
client.send(
    PostUpvotedEvent(
        data=PostUpvotedEventData(count="bad data"),
    ).to_event()
)
```

PR #920, opened 2024-09-17
Remove duplicate h1 search results

djfarrelly:
Currently, our Algolia results duplicate the h1 tags somehow and for key terms that match a docs page h1 partially or exactly like "throttle" o "local dev" the h1 is duplicated and the page shows up twice in the results. 

This change filters the results from Algolia on the client and removes any duplicates that link to the same exact h1 (or `lvl1` in Algolia's results).

### After
image

### Before
image


charlypoly (approved):
⚡

PR #951, opened 2024-10-03
Add link to pricing plan to concurrency.mdx

01programs:
It's useful to see right away what the given limits are per plan, so I'm adding a link.


charlypoly (approved):
Good point @01programs ⚡

PR #953, opened 2024-10-10
Fix title of invoke_by_id.mdx

01programs:
The title was just "invoke" and should be "invoke by id"


charlypoly (approved):
Thank you for spotting this @01programs! ⚡

PR #968, opened 2024-10-28
Add timeouts to docs

tonyhb:
This adds docs for the previously undocumented `timeouts` property.

### DIFF
--- a/pages/docs/features/inngest-functions/cancellation/cancel-on-timeouts.mdx
+++ b/pages/docs/features/inngest-functions/cancellation/cancel-on-timeouts.mdx
@@ -18,8 +18,6 @@ In the following examples, we'll explore how to configure the timeout property a
     { key: "go", title: "Go" },
   ]}>
 
-<GuideSection show="typescript">
-</GuideSection>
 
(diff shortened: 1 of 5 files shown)


pages/docs/features/inngest-functions/cancellation/cancel-on-timeouts.mdx
```diff
@@ -0,0 +1,129 @@
...
+
+<GuideSelector
+  options={[
+    { key: "typescript", title: "TypeScript" },
+    {/* { key: "python", title: "Python" }, */},
+    { key: "go", title: "Go" },
+  ]}>
+
+
+### `timeouts.start` - Adding timeouts to unstarted runs
```
charlypoly, line 22: @tonyhb, suggestion: we might want to wrap also the content surrounding the code samples in `<GuideSection>` as Go configuration properties have a different name. wdyt?

tonyhb: I thought about that, but didn't know if capitalization was such a big deal here.  I can, if we don't mind the repeated content.  Your call.

PR #976, opened 2024-10-30
Use CodeHike filenames. Refresh CTA using new design styles.

djfarrelly:
Using codehike file names:
Screen Shot 2024-10-30 at 4 07 01 PM

Refreshed CTA, using styles from recent landing page. The only CTA was using old brand colors.

image


charlypoly (approved):
Thank you! ✨ 💅🏼

PR #983, opened 2024-11-01
Link fixes, workflow engine page refresh

djfarrelly:


charlypoly (commented):
@djfarrelly, I've left comments on integrating the Workflow Kit in the Durable Workflows messaging.

PR #991, opened 2024-11-11
fix(tracking): re-render issue altering data + cleaning

charlypoly:
## Before 
Race condition on the `useAnonymousID()` hook is due to multiple renders, leading to a `first_touch` always being `false`.

Screenshot 2024-11-11 at 15 58 42

_Page loaded without anonID cookie is flagged as existing user_

## After

No more re-renders issues + UTM tracking on both first touch and all page views

Screenshot 2024-11-11 at 16 27 27

_No anonID cookie: properly flagged as first touched with dedicated UTM props_

Screenshot 2024-11-11 at 16 29 00

_Second visit: properly flagged with dedicated UTM props_

### DIFF
--- a/public/inngest-sdk.js
+++ b/public/inngest-sdk.js
@@ -154,6 +154,7 @@
     "utm_content",
     "utm_medium",
     "utm_term",
+    "ref",
   ];
 
   function context(eventData) {
--- a/public/inngest-sdk.js
+++ b/public/inngest-sdk.js
@@ -38,7 +38,7 @@
       return false;
     }
     event.data = event.data || {};
-    assign(event.data, context());
+    assign(event.data, context(event.data));
 
     // The event.user object should take precedence over the identify() attributes
     // called.  Copy the event user attributes into a new variable so that we can
@@ -73,7 +73,7 @@
   Inngest.track = function (eventName, data) {
     evt = {
       name: eventName,
-      data: assign(data || {}, context()),
+      data: assign(data || {}, context(data)),
       user: user,
     };
     Inngest.event(evt);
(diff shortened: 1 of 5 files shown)


public/inngest-sdk.js
```diff
@@ -148,7 +148,29 @@
...
+      const urlParams = new URLSearchParams(window.location.search);
+      // we track both first touch UTMs props and all touchpoints UTMs
+      for (let utmKey of utmKeys) {
+        utmProps[utmKey] = urlParams.get(utmKey);
+        if (isFirstTouch) {
+          utmProps[`first_${utmKey}`] = urlParams.get(utmKey);
+        }
+      }
+    } catch (e) {}
+
```
djfarrelly, line 173: This code looks duplicative of the code after defining `var data`  and this line. We should remove that other UTM parsing code as future changes to this code would likely be overwritten by the other code after this if statement:

```js
if (window && window.location.search.length) {
```

shared/trackingHooks.tsx
```diff
@@ -13,32 +13,39 @@ export const abExperiments = {
...
-    const sixMonthsFromNow = new Date(Date.now() + 6 * 30 * 24 * 60 * 60 * 1000);
-    setAnonymousID(newAnonymousID, { domain: process.env.NEXT_PUBLIC_HOSTNAME, path: "/", expires: sixMonthsFromNow, sameSite: 'lax' });
-    return {
-      anonymousID: newAnonymousID,
-      existing: false,
-    };
+      cached.current = {
+        anonymousID: newAnonymousID,
+        existing: false,
+      };
```
djfarrelly, line 40: Does this just help when this hook is called multiple times on a single page render so it doesn't perform this multiple times and overwrite the value? 

How does this work with page loads/transitions? does existing change to true?

djfarrelly (approved):
@charlypoly - I'm approving the code pending the removal of the duplicate code mentioned in the comments

shared/trackingHooks.tsx
charlypoly: > Does this just help when this hook is called multiple times on a single page render so it doesn't perform this multiple times and overwrite the value?

Exactly, this avoids the race condition of the hook being called multiple times when rendering the page at the initial load.
Without the use of this `cached` ref, the hook is called multiple times in a row, ending in never flagging a user as new.


> How does this work with page loads/transitions? does existing change to true?
The client-side routing changes are handled here, I believe: https://github.com/inngest/website/blob/cd6bdf76519623f6de9c75e3656e8eea8448e76e/pages/_app.tsx#L96

public/inngest-sdk.js
charlypoly: Good catch! 🦅 
Solved in [72508fb](https://github.com/inngest/website/pull/991/commits/72508fbc7c6303ec883d0717dac9c402f1ee47c6)

djfarrelly (approved):
LGTM

PR #993, opened 2024-11-14
Fix Sentry middleware setup example for Python

rbm:
Inngest expects a list of uninitialized middleware here, i.e. the actual SentryMiddleware class rather than an instance of it.

The class is initialized in MiddlewareManager.add when sending events.


rbm:
The code in the example currently throws an exception because `SentryMiddleware` has required parameters (which `MiddlewareManager` supplies)

charlypoly (approved):
Thank you for spotting this @rbm! ⚡

PR #995, opened 2024-11-15
Add missing class attribute type in pydantic.mdx

01programs:
The subclass attribute 'name' also needs to be of type 'ClassVar' otherwise you get and Attribute error for 'name' when doing ' trigger=inngest.TriggerEvent(event=PostUpvotedEvent.name)' (the class is not instantiated here)


amh4r:
Thank you for the fix!

amh4r (approved):

PR #1008, opened 2024-11-25
Add sub heads for python examples in step-parallelism.mdx

01programs:
Added sub headlines for the python (a)sync examples because I thought these are the same code block twice and only saw after using a diff checker. 

Note: There is a randering issue on GH realted to the '<CodeGroup>' tag. I can not tell if that's just GH or also the website after merge!


charlypoly (approved):
Thank you for your contribution @01programs! ⚡

PR #1020, opened 2024-12-08
docs: sidebar editor empty when following user guides

JacobReynolds:
The sidebar component was blank because it was rendering the fragment provided to it, instead of the built-in functionality.


charlypoly (approved):
Good catch @JacobReynolds 🦅 
I forgot to update this example, thank you!

PR #1033, opened 2024-12-30
Update nextjs-quick-start.mdx

jakedee:
Fix example route handler code


charlypoly (approved):
Thank you for spotting this @jakedee! ⚡

PR #1038, opened 2025-01-08
Update status widget to use incident.io

djfarrelly:
Change to the new API now that we've migrated


charlypoly (approved):
I restarted the failed build and reviewed the preview, it looks great! ⚡

PR #1047, opened 2025-01-20
Added guide for disabling path discovery

chrisidakwo:
See https://github.com/inngest/inngest/issues/1298

The console is typically an important tool of the development experience, and devs should be able to easily get information on how to opt out of a feature that constantly logs route discovery information that is not exactly needed for most of the development time.


djfarrelly:
hi @chrisidakwo - thanks for this contribution, to make it more searchable for users beyond the quick start, we've added documentation in the dev server docs and the FAQ: https://github.com/inngest/website/pull/1048

We'll review your recommended changes and determine the best way to incorporate both your ideas and the changes in #1048. In addition, we're planning to change how auto-discovery work in an upcoming dev server release so this isn't as annoying to users.

chrisidakwo:
@djfarrelly Nice! Great product too!

PR #1053, opened 2025-01-27
feat(banner): link to the upcoming Inngest + Render webinar

charlypoly:


app/layout.tsx
```diff
@@ -56,8 +56,9 @@ export default function RootLayout({
         </Suspense>
       </head>
       <body className="dark font-sans">
-        <AnnouncementBanner href="/blog/ai-orchestration-with-agentkit-step-ai?ref=banner">
-          Introducing AgentKit and step.ai: orchestrating AI with confidence
+        <AnnouncementBanner href="https://www.linkedin.com/events/buildahackernewsaiagentwithinng7287214928480874496/theater/">
```
djfarrelly, line 59: Do these link support UTMs? Any sort of data on this

djfarrelly (approved):
This will only show on app router pages, I think we may need this to show on old pages router pages (if you want): pages/_app.tsx

app/layout.tsx
charlypoly: I don't know :/

djfarrelly: We could use a short link if we want

charlypoly:
```suggestion
        <AnnouncementBanner href="https://innge.st/ai-agent-webinar">
```

PR #1057, opened 2025-01-30
feat(docs): add LLM docs generator script

joelhooks:
Adds a TypeScript script to generate a single text file containing all MDX documentation, cleaned and formatted for LLM consumption. The script:

- Recursively processes all MDX files in the docs directory
- Removes MDX-specific content (imports, exports, JSX)
- Preserves document structure and metadata
- Converts components and links to readable text
- Outputs to public/llms.txt


djfarrelly:
From quick review this looks great. We may want to update this to `llms-full.txt` as that seams to be the pattern whereas `llms.txt` is the shorter version with links. 

Also, we might want to adjust the parsing so we handle <Resource/> and similar better. From comparing things with MCP, they do leave some MDX in there so I'm not sure how much we need to rip out or just cleanup: https://modelcontextprotocol.io/llms-full.txt

We'll aim to sort this out and get it merged in the next few days!

Thanks @joelhooks!

joelhooks:
sweet, i wanted this for my own local setup so figured i'd kick it over the fence and those changes sound great

it's very useful in practice!

could also argue that breaking of agentkit etc might make sense but i like the full context and it seems to function well

charlypoly (approved):

djfarrelly:
Thanks @joelhooks! These changes were merged part of #1088 and should be live in a few minutes.

PR #1080, opened 2025-03-13
docs: Fix data type and description

jakedee:
Fix "period" data type and description.

Should this data type be more explicit than `string`?
I'm not sure what the data type actually is, but my understanding is that it can only handle specific "duration" strings e.g. 5s, 10m, 4h, 6d etc.


djfarrelly (approved):
Thanks @jakedee! We appreciate the fix!

PR #1102, opened 2025-04-14
fix: bun install global

kadumedim:
`bun add global` is wrong and it would attempt to install a package named 'global' alongside 'inngest-cli' rather than installing 'inngest-cli' globally."

It should be `bun install --global inngest-cli@latest`


charlypoly:
Thank you for spotting this @kadumedim! ⚡

PR #1103, opened 2025-04-22
Add docs on connect Syncing and rollbacks

BrunoScheufler:


charlypoly (approved):

pages/docs/setup/connect.mdx
```diff
@@ -367,6 +367,14 @@ This view is helpful for debugging connection issues or verifying rolling deploy
 
 ![App worker observability](/assets/docs/connect/cloud-app-workers.png)
 
+## Syncing and Rollbacks
+
+When deploying a new version of your application, the first instance to reach `connect()` will kick off a sync during the connection setup process. This will update function configurations to the desired state configured in your code.
```
djfarrelly, line 372:
```suggestion
When you deploy a new version of your application, the first worker to connect to Inngest will automatically sync your app. This will update function configurations to the desired state configured in your code.
```

Using some terms from this and other docs: "sync your app" and "worker" - we could also link the lifecycle section which mentions this as well.

pages/docs/setup/connect.mdx
```diff
@@ -367,6 +367,14 @@ This view is helpful for debugging connection issues or verifying rolling deploy
...
+
+`connect` supports rolling releases: During a deployment of your app, Inngest will run functions on all connected workers, regardless of the version, as long as they are able to process a request for a given function. This prevents traffic from concentrating on a single instance during rollouts and causing a thundering herd issue.
```
djfarrelly, line 374: When reading this my key question is what "version" are we talking about here? An "app version?" or a function version? function versions are opaque to the user so we may want to clarify this.

If I understand correctly, as a rolling release happens workers running old versions of my app might still handle function execution which seems more like a warning rather than a feature. We may want to phrase it like that. Not sure.

pages/docs/setup/connect.mdx
```diff
@@ -367,6 +367,14 @@ This view is helpful for debugging connection issues or verifying rolling deploy
...
+
+Once all old workers have terminated after a deployment, you can roll back to an old version by bringing back an old worker. Similar to the deployment process, this will update the function configuration to the previous state and gradually allow you to shift traffic to the old version by bringing up more old workers while terminating workers running the newer version.
```
djfarrelly, line 376: While this makes sense to me, some sort of diagram even ascii may be helpful for clarity. As I read this, this mostly means "Once a new version has been rolled out to 100% of your workers, deploying an old version of your app will begin a rollback. The old version will be synced will again be the default as the old version rolls out to 100% of your workers."

djfarrelly (approved):
Approved with some suggestions/questions/thoughts - feel free to take what you want and ship it!

pages/docs/setup/connect.mdx
BrunoScheufler: Agree, do we have a nice way to include diagrams or should this be a collaboration with design?

PR #1115, opened 2025-05-09
chore: fixed overflow dark mode

jakobevangelista:
### Summary
Currently in the docs, if you scroll too far down, you can see the canvas underneath which is default light mode color

Screenshot 2025-05-09 at 1 10 12 PM

This fixes the unintended behavior

Screenshot 2025-05-09 at 1 10 28 PM

Add dark-mode rules to `styles/globals.css` to fix overflow

### DIFF
--- a/shared/Docs/Layout.tsx
+++ b/shared/Docs/Layout.tsx
@@ -146,6 +146,19 @@ const modeScript = `
 
   document.documentElement.classList.add('docs');
 
+  const styleEl = document.createElement('style');
+  styleEl.id = 'dark-mode-styles';
+  styleEl.textContent = \`
+    html.dark {
+      background-color: rgb(var(--color-carbon-1000)) !important;
+      min-height: 100vh;
+    }
+    html.dark body {
+      background-color: rgb(var(--color-carbon-1000)) !important;
+    }
+  \`;
+  document.head.appendChild(styleEl);
+
   // change to "let = darkModeMediaQuery" if/when this moves to the _document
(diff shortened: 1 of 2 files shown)


shared/Docs/Layout.tsx
```diff
@@ -146,6 +146,19 @@ const modeScript = `
...
+  styleEl.textContent = \`
+    html.dark {
+      background-color: rgb(var(--color-carbon-1000)) !important;
+      min-height: 100vh;
+    }
+    html.dark body {
+      background-color: rgb(var(--color-carbon-1000)) !important;
+    }
+  \`;
+  document.head.appendChild(styleEl);
```
djfarrelly, line 160: **question**: Hey @jakobevangelista, instead of dynamically adding this via JS, can we inline this in the _document.tsx or similar? My only concern would be debugging this in the future and not sure where the style is set.

djfarrelly (commented):
I am curious why it doesn't work in global.css (there is a gotcha - we have 2 with the app router setup (https://github.com/inngest/website/blob/main/styles/globals.css, https://github.com/inngest/website/blob/main/app/globals.css).

I'm OK with merging this to fix it, but am curious if there is another way to do this other than inject via JS which might cause debugging headaches in the future for other devs.

shared/Docs/Layout.tsx
jakobevangelista: Updated with [this](https://github.com/inngest/website/pull/1115/commits/62c76a3ec7fe82c8961995d288d1705e5519e20d) commit

djfarrelly (approved):

PR #1124, opened 2025-05-20
Added platform popover

jakobevangelista:
Added platform popover for the header

Small block because I don't know where each item is linked to


jakobevangelista:
Combined in [#1131](https://github.com/inngest/website/pull/1131)

PR #1177, opened 2025-06-26
Extended calculator ranges and fixed pricing calculations

jakobevangelista:


components/RedesignedPricing/PricingCalculator.tsx
```diff
@@ -173,17 +195,17 @@ function PlanSummary({ plan, price }: { plan: Plan; price: string }) {
 
 export function PricingCalculatorPage() {
   const [runs, setRuns] = useState(50000);
-  const [steps, setSteps] = useState(20);
+  const [steps, setSteps] = useState(3);
```
djfarrelly, line 198: Why is this 3? Should we just make this 50000?

djfarrelly (approved):

components/RedesignedPricing/PricingCalculator.tsx
charlypoly: @jakobevangelista, following up on this 👆🏼

PR #1183, opened 2025-07-04
Load code block content from file

amh4r:
Add support for loading code block content from a file. So you can do something like this:
image

And it'll render:
image

Using this Python file:
image


charlypoly (commented):
@amh4r, did it work well with the local hot-reloading? Ex: updating a code file is taken into account if we refresh the associated doc page.

PR #1184, opened 2025-07-04
Semicolon in wrong place

kadumedim:
Just a wrong semicolon, should be a comma instead :)


djfarrelly (approved):
Thanks for catching that @kadumedim!

PR #1250, opened 2025-09-05
Fix broken links

djfarrelly:
See: https://github.com/inngest/website/actions/runs/17499205661/job/49708053586


Benanna2019 (approved):
That was fast!

PR #1258, opened 2025-09-11
Use compiled go snippets

lkasinathan:
Use compiled go snippets on the website to prevent the documentation on website from going stale.

Migrating most of the go code snippets on the website to snippets/go/docs which will be compiled with the latest inngestgo SDK.


snippets/go/Makefile
```diff
@@ -2,3 +2,4 @@
 check:
 	cd v0_8 && go build -o /dev/null
 	cd v0_11 && go build -o /dev/null
+	cd docs && go get -u github.com/inngest/inngestgo && go build -o /dev/null
```
amh4r, line 5: Is the `go get` necessary since `github.com/inngest/inngestgo` is already in `docs/go.mod` 🤔

lkasinathan: Not strictly necessary, but I wanted the check to run against the latest SDK so I added a `go get -u` here. lmk if there's a better way to do this!

PR #1281, opened 2025-09-25
Updated Realtime doc to include Python and added Subscribe page

Benanna2019:
I went ahead and made subscribe its own page. I feel that could be confusing to people and having it explicitly clear that subscribe is a new step I think can help out as people start getting setup.

### DIFF
--- a/pages/docs/features/realtime.mdx
+++ b/pages/docs/features/realtime.mdx
@@ -18,7 +18,7 @@ Realtime user experience is a core requirement for any web application, especial
 
 ## Concepts
 
-There are two core parts of Realtime: **publishing** and **[subscribing](/docs/features/realtime/subscribe)**. You **publish** data from your functions and **subscribe** to data in your application, either browser or server.
+There are two core parts of Realtime: **[publishing](#publishing)** and **[subscribing](#subscribing)**. You **publish** data from your functions and **subscribe** to data in your application, either browser or server.
 
 Publishing data is done using the `publish()` function and has three components:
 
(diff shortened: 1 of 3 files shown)


pages/docs/features/realtime.mdx
```diff
@@ -77,8 +99,25 @@ export const inngest = new Inngest({
...
+```py {{ title: "client.py" }}
+import logging
+
+import inngest
 
+logger = logging.getLogger("uvicorn.inngest")
+logger.setLevel(logging.DEBUG)
+
+inngest_client = inngest.Inngest(app_id="fast_api_example", logger=logger)
+```
```
djfarrelly, line 115: I don't think we need to cover the Python client here as that should be a pre-requesite. The only reason we show the client in the TS SDK example is b/c it requires middleware whereas Python has realtime built in, just needs to import.

djfarrelly: Reading onward, I see why you did this, b/c the client is used later. Perhaps we can instead link to another doc that says "You'll need to have set tup the client which you would have learned in our Python quick start guide" - or something like that.

pages/docs/features/realtime/subscribe.mdx
```diff
@@ -0,0 +1,146 @@
...
+      const token = await getSubscriptionToken(inngest, {
+        channel: `user:${userId}`,
+        topics: ["ai"],
+      })
+
+      res.json({ token })
+    })
+    ```
+
+    ```py {{ title: "Python" }}
```
djfarrelly, line 64:
```suggestion
    ```py {{ title: "Python - Fast API" }}
```

pages/docs/features/realtime.mdx
Benanna2019: Yeah, I think this makes sense. I think I'll role the subscribe doc back into this and then next phase of realtime docs will add more explanation around subscribing and so on so I can go into detail on sdk's and subscribing, clients and so on.

PR #1507, opened 2026-03-30
Mark Durable Endpoints as developer preview

amh4r:


Benanna2019 (approved):

djfarrelly:
We're deciding to keep this in beta. We should instead, clearly list out any limitations of the feature currently, like lack of post body. Closing this PR.

PR #1510, opened 2026-03-31
Restructure step metadata docs into reference page

Benanna2019:
## Summary

Replaces the single step metadata page from #1498 with a focused **reference** page following the Diataxis framework. Strips narrative and how-to content, keeping just the API surface:

- `metadataMiddleware()` setup
- `step.metadata()` and `inngest.metadata` method tables
- Scoping rules and auto-detection behavior
- Custom kinds and built-in metadata kinds
- Size limits and error codes
- Limitations

The companion how-to page is in a separate PR.

Relates to DEV-276.

## Requested reviewers

@AndyInternet @djfarrelly


djfarrelly (changes requested):
key high level notes and some other questions for if some of this info could be useful elsewhere - generally I like the direction and the split here between this and the guide.

djfarrelly (approved):
Approving with some open questions from the previous review that are non-blocking. The how it works might be for here or the guide

AndyInternet (approved):
Need to fix import. Everything else looks great!

PR #1511, opened 2026-03-31
Add step metadata how-to guide

Benanna2019:
## Summary

Adds a practical **how-to** page for step metadata following the Diataxis framework. Each section is a discrete task with code and just enough context inline:

- Attach metadata to a step (both `step.metadata` and `inngest.metadata` inside `step.run`)
- Use custom kinds to group metadata
- Batch multiple updates in one step
- Scope metadata to a specific target
- Update metadata on a different run
- Update metadata from outside a function
- View metadata in the dashboard

Explanation content (batched vs API path, durability) is folded inline as context rather than a standalone section.

Companion to the reference page PR. Relates to DEV-276.

## Requested reviewers

@AndyInternet @djfarrelly

### DIFF
--- a/pages/docs/features/inngest-functions/steps-workflows/step-metadata-how-to.mdx
+++ b/pages/docs/features/inngest-functions/steps-workflows/step-metadata-how-to.mdx
@@ -1,12 +1,14 @@
 import { CodeGroup, VersionBadge, Info, Warning } from "src/shared/Docs/mdx";
 
-export const description = "Practical guides for attaching metadata to steps and runs, using custom kinds, batching updates, and working with cross-run metadata.";
+export const description = "Practical guides for attaching metadata to function runs, using custom kinds, batching updates, and working with cross-run metadata.";
 
-# Step metadata how-to <VersionBadge version="TypeScript only" />
+# Metadata how-to <VersionBadge version="TypeScript only" />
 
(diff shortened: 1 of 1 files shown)


pages/docs/features/inngest-functions/steps-workflows/step-metadata-how-to.mdx
```diff
@@ -0,0 +1,157 @@
+import { CodeGroup, VersionBadge, Info, Warning } from "src/shared/Docs/mdx";
+
+export const description = "Practical guides for attaching metadata to steps and runs, using custom kinds, batching updates, and working with cross-run metadata.";
+
+# Step metadata how-to <VersionBadge version="TypeScript only" />
```
djfarrelly: note: just an open question before this is added to the public docs if we call this "run metadata" "metadata" or "step metadata" - "step metadata" is our internal name and this feature more attaches metadata to runs, not steps.

djfarrelly (changes requested):
Hey @Benanna2019 - I like a lot of parts about this and I actually think this could be the default guide in the main "Learn" side of the docs and the other guide should be in the "Reference" tab. 

With that in mind, I'd love for the intro to kind of frame the problem a little bit more. I think with just a little bit of "what this is for" / "what problem am do I need solved" - this guide would hit home really hard for folks. There's lots to like here, but I think we could get it to "click" for readers a lot faster then the examples in this flow get them there really fast which is great for a how-to.

djfarrelly (approved):

AndyInternet (approved):

PR #1525, opened 2026-04-08
pricing: show X for Dedicated slack channel on Pro tier

LCraigie:
Pro previously displayed $200; use the same not-included icon as Basic.

Made-with: Cursor


patrickholcomb9 (approved):
P0 support updates looks good!!

PR #1528, opened 2026-04-13
docs: correct TypeScript SDK documentation

Linell:
Update the documentation to correctly reflect the updates we've made to the SDK.

Related:
* https://github.com/inngest/inngest-js/issues/1452

### DIFF
--- a/pages/docs/typescript.mdx
+++ b/pages/docs/typescript.mdx
@@ -27,10 +27,8 @@ This comes with powerful inference; we autocomplete your event names when select
 import { eventType, Inngest, staticSchema } from "inngest";
 
 type UserSignup = {
-  data: {
-    email: string;
-    name: string;
-  };
+  email: string;
+  name: string;
 };
 
 export const userNewSignup = eventType("user/new.signup", {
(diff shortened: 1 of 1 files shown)


pages/docs/typescript.mdx
```diff
@@ -65,20 +63,19 @@ Here's an example of sending an event within a Next.js API handler:
...
-    data: {
-      email: user.email,
-      name: user.name,
-    }
+  const userNewSignupEvent = userNewSignup.create({
+    email: user.email,
+    name: user.name,
   });
+
+  await inngest.send(userNewSignupEvent);
```
amh4r: Should we inline the `create` call (rather than creating a new variable)? Might make it read a little cleaner, and more similar to how users would write code

pages/docs/typescript.mdx
```diff
@@ -130,10 +123,10 @@ export default inngest.createFunction(
     const accountSetupCompleted = await step.waitForEvent(
       "wait-for-setup-complete",
       {
-        event: "user/account.setup.completed",
+        event: userAccountSetupCompleted,
         timeout: "2d",
         // ⬇️ This matches both events using the same property
-        // Since both events types are registered above, this is match is typesafe
+        // Passing the `EventType` variable makes the returned event data typesafe
```
amh4r: I think we removed the type safety on this (even before the v4 project). Can you confirm if that's true?

Benanna2019:
Linear ticket: DEV-301

This addresses the same issues tracked there (schema nesting, async handler, eventType.create() pattern). Ready for review.

Benanna2019 (approved):

PR #1534, opened 2026-04-14
Add known limitation for waitForEvent in clause stripping

Benanna2019:
## DEV-304

Adds a warning callout to the `step.waitForEvent()` documentation about a known limitation where the `in` operator in CEL expressions may be silently stripped under certain conditions.

### Issue
When using the `if` option with CEL expressions containing the `in` operator (e.g., `async.data.function_id in ["a", "b"]`), the operator may be silently stripped from the expression, causing `waitForEvent` to match more broadly than intended.

### Solution
Documented the issue and provided a workaround using multiple equality checks with `||` operator.

### Related
- GitHub issue: https://github.com/inngest/inngest/issues/3907


pages/docs/reference/typescript/v4/functions/step-wait-for-event.mdx
```diff
@@ -71,4 +71,15 @@ Use `step.waitForEvent()` to pause your function's execution until a matching ev
 
 <Callout>
   `step.waitForEvent()` must be called using `await` or some other Promise handler to ensure your function sleeps correctly.
-</Callout>
\ No newline at end of file
+</Callout>
+
+<Callout variant="warning">
+  **Known limitation with `in` clauses:** When using the `if` option with CEL expressions, the `in` operator (for example, `async.data.function_id in ["a", "b"]`) may be silently stripped from the expression under certain conditions, causing your `waitForEvent` to match more broadly than intended. As a workaround, use multiple equality checks with `||` instead:
```
Linell: Would it be better to wholesale say it's not supported, instead of framing it as a workaround? 🤔

Linell (approved):

pages/docs/reference/typescript/v4/functions/step-wait-for-event.mdx
Benanna2019: I think we can say its not supported. If someone needs a work around we can keep this in the back pocket or something. Or just negate it all together.

djfarrelly: Hey @Benanna2019 - I think Linell has a point in terms of language. We should phrase it more directly, as in "
```suggestion
  **No `in` operator support. Use multiple equality checks with `||` instead.** CEL's `in` operator is not supported with `if` expressions, instead, use multiple equality checks (`==`) with an or operator (`||`) to achieve the same result.
```

pages/docs/reference/typescript/v4/functions/step-wait-for-event.mdx
```diff
@@ -71,4 +71,15 @@ Use `step.waitForEvent()` to pause your function's execution until a matching ev
...
+
+  {`// Instead of:
+  // async.data.function_id in ["app-test-a", "app-test-b"]
+  // Use:
+  (async.data.function_id == "app-test-a" || async.data.function_id == "app-test-b")`}
```
djfarrelly:
```suggestion
  {`// Do not use this:
  // async.data.function_id in ["app-test-a", "app-test-b"]
  // Use this:
  (async.data.function_id == "app-test-a" || async.data.function_id == "app-test-b")`}
```

djfarrelly (changes requested):
Hey @Benanna2019 - leaving my thoughts on the copy here!

PR #1539, opened 2026-04-16
Add guest blog: self-improving agent (Mitchell Alderson)

LCraigie:
Made-with: Cursor


LCraigie:
@djfarrelly Cursor was having trouble replacing the repo link in here, but it should be https://github.com/inngest/inngest-self-learning-agent instead of Mitch's repo at the end. How do I change code from inside GitHub?

djfarrelly (approved):

PR #1543, opened 2026-04-20
Document optional jitter field for cron triggers

dlambrig:
Add jitter documentation to the scheduled functions guide with examples. Update the TypeScript trigger reference with the cron trigger object form and jitter property.

Note- There isn’t currently a Go reference page in this repo parallel to the TypeScript and Python trigger docs, (i.e.  there wasn’t an equivalent place to add a Go-specific jitter blurb.)

### DIFF
--- a/pages/docs/guides/scheduled-functions.mdx
+++ b/pages/docs/guides/scheduled-functions.mdx
@@ -247,3 +247,58 @@ On the free plan, if your function fails 20 times consecutively it will automati
 
 Inngest's cron behavior follows the underlying cron library and does not apply special DST correction. To reduce risk, avoid transition-hour schedules (such as `2:00 AM` in many US regions, or `12:00 AM` in some other regions), and prefer `TZ=UTC` when you need consistent execution timing.
 </Callout>
+
+## Adding jitter
+
+By default, cron functions fire at the exact scheduled time. If you have many cron functions on the same schedule, they all fire at once which can create load spikes on your system or on third-party APIs.
+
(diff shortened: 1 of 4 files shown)


pages/docs/guides/scheduled-functions.mdx
```diff
@@ -246,4 +246,59 @@ On the free plan, if your function fails 20 times consecutively it will automati
...
 Inngest's cron behavior follows the underlying cron library and does not apply special DST correction. To reduce risk, avoid transition-hour schedules (such as `2:00 AM` in many US regions, or `12:00 AM` in some other regions), and prefer `TZ=UTC` when you need consistent execution timing.
-</Callout>
\ No newline at end of file
+</Callout>
+
+## Adding jitter
+
+By default, cron functions fire at the exact scheduled time. If you have many cron functions on the same schedule, they all fire at once which can create load spikes on your system or on third-party APIs.
+
+You can add an optional `jitter` to spread out the execution. When jitter is set, each cron occurrence fires at a random time within the jitter window after the scheduled boundary. The jitter is deterministic per occurrence — retries and restarts produce the same delay.
```
lkasinathan: > The jitter is deterministic per occurrence — retries and restarts produce the same delay.

I suggest removing this sentence. The jitter is deterministic across retries of the _internal_ cron job and only applies to the first attempt of the user's job. If the user job is retried there is no additional jitter on retries and we just retry immediately.

lkasinathan (approved):
Left one suggestion, other than that, looks great!

PR #1557, opened 2026-04-28
Refresh Outtake customer case study page copy.

LCraigie:
Update the Outtake customer page content and section wording while keeping changes scoped to the customer page only.

Made-with: Cursor


patrickholcomb9 (approved):
Updates look great! 

One small formatting item - not a blocker: do we want the customer quotes to be in "Block Quote" format (like they are on [current live post](https://www.inngest.com/customers/outtake))? 

And the first quote on this PR ("“Our pipelines are intricate, with many interdependent components..") is not italicized, but the other quotes are. I'm ok with either (or Block Quote), but would be good to have them consistent!

PR #1558, opened 2026-04-28
Add Outtake customer story card to blog feed.

LCraigie:
Create a redirect-style blog card with featured image and focus it as the top blog hero linking to the Outtake customer page.

Made-with: Cursor


patrickholcomb9 (approved):
looks great, lgtm!

PR #1559, opened 2026-04-28
Update Outtake preview title and OG image

LCraigie:
Made-with: Cursor


patrickholcomb9 (commented):
One tweak: title on card reads "How Outtake uses Inngest to ensure the reliability of millions of thteat scans every minute" >> change "thteat" to "threat"

patrickholcomb9 (approved):
lgtm!

PR #1561, opened 2026-04-29
feat: content download form + reusable [asset] route framework

patrickholcomb9:
[Adds gated content download landing pages built as a reusable framework so new content downloads ship via a CONTENT_ASSETS lookup entry + PDF, with no new page code required. The dynamic /content/[asset] route renders any registered asset; bespoke per-asset designs can override at app/content/<slug>/page.tsx (Next.js resolves static routes first).

Form behavior is centralized in <ContentDownloadForm>, which wraps ContactForm with content-download-specific event params so every asset fires the same Segment Track event with an identical payload shape.

First-touch attribution (utm_*, ref, landing URL, referrer) is captured on first visit via a 6-month cookie and replayed on every form submission, so attribution survives the gap between a UTM-tagged first landing and a later form submission from a static URL.

Two adjacent fixes that came up during this work:
- Extract FORM_TYPE constants out of the "use client" ContactForm module. Server Components on /contact, /yc, /landing/ai, /compare-to-temporal were silently resolving the import to undefined, breaking GTM event names. New ContactForm.constants.ts has no "use client" directive and is safe for both Server and Client Components.
- Scope Segment cdnURL to production only. The earlier merge fixed apiHost for local dev but missed cdnURL, which was hardcoded to analytics-cdn.inngest.com and blocked the local dev Segment loader.

The Segment->Customer.io and Segment->Attio destination mappings (which handle CRM person upserts and timeline events) are configured separately in the Segment workspace and end-to-end tested in production. No website-side code is required for those integrations to work.](feat: content download form + reusable [asset] route framework)


components/ContactForm.tsx
```diff
@@ -92,66 +137,139 @@ export default function ContactForm({
...
+          className
+        )}
+      >
+        <h2 className="text-xl font-semibold text-white">
+          {isContentDownload ? "You're all set" : "Thanks — we got it"}
+        </h2>
+        <p className="text-basis">
+          {successMessage ||
+            (isContentDownload
+              ? "Check your email — the content is on its way. If you don't see it in a few minutes, check your spam folder."
```
djfarrelly: It looks like we're redirecting them after anyway, so would we also send them an email or just redirect them to a page that they can download the assets?

djfarrelly (approved):

PR #1562, opened 2026-04-30
Refresh patterns page for AI orchestration focus

Benanna2019:
DEV-316

## Summary

Refreshes the `/patterns` page to align with Inngest's AI orchestration positioning. Inspired by the Analogue case study and conversations about bringing patterns back into the site nav.

**New pattern:**
- **Flash Sales and Bursty Workflows** — Covers throttle, concurrency, debounce, and idempotency with production code examples from the Analogue case study

**5 existing patterns refreshed** with AI orchestration framing (LLM offloading, multi-step AI pipelines, parallel model calls, model callbacks, human-in-the-loop approval flows):
- Keeping your API fast
- Build reliable webhooks
- Reliably run critical workflows
- Running functions in parallel
- Building flows for lost customers

**4 patterns removed from index** (files kept, just not linked):
- Running code on a schedule
- Running at specific times
- Cancelling scheduled functions
- Batching jobs via fan-out

**Page structure reorganized** into three sections:
1. Flow Control
2. Durable Workflows
3. Event Coordination

## Context

- Lauren's Analogue case study provided the source material for the new bursty workflows pattern
- Ben proposed adding Patterns to site nav (Apr 16, #dev-rel-docs); Dan confirmed the existing page is still live
- Lauren and Ben discussed architectural patterns as content for onboarding guides (Apr 17, #marketing-team)

## Test plan

- [ ] Verify `/patterns` index renders with new sections and cards
- [ ] Verify each pattern page renders correctly (especially the new flash-sales one with the table)
- [ ] Confirm removed patterns still render if accessed directly via URL (files not deleted)
- [ ] Check code examples render properly with syntax highlighting


pages/patterns/index.tsx
```diff
@@ -1,223 +1,173 @@
...
+      </Container>
+
+      {/* Featured pattern */}
+      {featuredPayload && (
+        <Featured
+          featured={featuredPayload.featured}
+          section={featuredPayload.section}
+          pattern={featuredPayload.pattern}
+        />
+      )}
```
SterlingChin: This render path is Hero → Featured → Footer — `sections` is built in `getStaticProps` but never iterated here. As a result only the featured pattern is reachable from /patterns; the other 7 work only via direct URL. Was the section grid pulled out intentionally, or still pending?

SterlingChin (commented):
Hey Ben — went deep on this one since Patterns is a big surface for the AI orchestration positioning. Inline comments cover the line-level stuff. Adding this top-level note for things that don't anchor cleanly to one line, plus an Agent Experience (AX) review since the `/md` endpoints and `?view=agent` toggle are real machine-readable infrastructure worth getting right.

## PR description vs. diff mismatches

A few discrepancies worth reconciling:

- **Body:** *"4 patterns removed from index (files kept, just not linked)."* **Diff:** `cancelling-scheduled-functions.mdx` and `running-code-on-a-schedule.mdx` are actually **deleted**, not just unlinked.
- **Body** says "Running at specific times" was removed — but it's still in the repo (`running-at-specific-times.mdx` with `pattern: schedule`) and will render into the Scheduling section.
- **Body** mentions "Batching jobs via fan-out" as removed — no file by that name appears in this diff. May have been deleted in a previous PR; worth confirming.
- **Body:** *"Page structure reorganized into three sections."* `patternsData.ts` defines 5 — see inline.

## Unrelated changes that look accidental

Diffs in this PR that aren't patterns work and aren't called out in the description. These look like they may have ridden in from a stale base:

- `useFirstTouch` removed from `pages/_app.tsx` (inline)
- `images.qualities: [75, 95]` removed from `next.config.mjs` (inline)
- Permanent redirect for `/2026-durable-execution-report` → `/content/ai-in-production-report-2026` removed from `next.config.mjs` (around the redirects block)
- `Event payload format` link dropped from docs sidebar in `navigationStructure.ts` (inline)

Worth a rebase against latest main and a pass to confirm each was intentional.

## Side note — pre-existing typo

`shared/Footer/FooterCallout.tsx:16` reads `"Ship reliable code without manage queues, infrastructure, or state"` → `manage` should be `managing`. Visible at the bottom of both patterns pages. Not your PR, but it shows up in the new UX so worth fixing while we're here.

---

## Agent Experience (AX) review

The `/patterns/md` and `/patterns/{slug}/md` endpoints and the `?view=agent` toggle are real machine-readable infrastructure — this is exactly the kind of work I want to see more of. Strong foundation. Comments below are about closing the discovery gap and a few content-level details.

### What's working

- `/patterns/md` and `/patterns/{slug}/md` return `text/markdown; charset=utf-8` with sensible cache headers (`max-age=300, s-maxage=3600`).
- Clean URL rewrites in `next.config.mjs` (`/patterns/md` → `/api/patterns/md`) so agents get a canonical URL, not `/api/...`.
- `markdown.ts` builds well-structured output: H1, blockquote subtitle, `URL:` and `Tags:` metadata lines, featured pattern first, then ToC, then sections. The comment *"agents pulling fresh context see the newest material first"* shows the right instinct.
- Humans/Agents toggle with H/M keyboard shortcuts is a great DX/AX education feature.

### The big AX gap: no discovery layer

An agent crawling `/patterns/flash-sales-and-bursty-workflows` currently has **no way to know** that the `.md` alternate exists. The page needs:

1. **`<link rel="alternate" type="text/markdown" href="/patterns/{slug}/md">`** in the HTML `<head>`. Standard convention, recognized by Anthropic, OpenAI, and most agent crawlers.
2. **HTTP `Link` header** with the same alternate, for clients that don't parse HTML.
3. **`/llms.txt` at the site root** (emerging convention from late 2024) listing the canonical content and linking to `/patterns/md` plus the existing `/docs-markdown/*` rewrites. ~1 day of work, outsized AX upside.
4. **Sitemap entries** for the `.md` alternates (`next-sitemap` supports per-URL `xhtml:link rel="alternate"`).

Without these, the `/md` endpoints are functionally invisible to anything that didn't already know the convention.

### Other AX touch-ups

- **`patternMarkdown` keeps relative doc links.** `body.trim()` is appended verbatim, so `/docs/...?ref=patterns` stays relative. An agent that copies the `.md` off-platform loses the base URL. Fix: rewrite leading-`/` paths to `https://www.inngest.com/...`, or include a `Canonical-Base:` header line agents can resolve against.
- **No freshness signal.** No `Last-Modified` / `Updated` line in the markdown output. Agents that cache need a refresh signal. Easiest path: thread MDX file mtime or a frontmatter `updatedAt` through.
- **Tracking won't survive copy/paste.** `?ref=patterns` on doc links is human-attribution. Consider `?ref=patterns-md` on the markdown-served variant so agent-driven traffic is measurable downstream.
- **MDX leakage risk.** `patternMarkdown` serves the body verbatim. Today all 8 patterns are clean markdown + fenced code (the `import` lines I see are inside TS code blocks). But there's no guard: if someone later adds `<Callout>` or `<CodeGroup>`, the `.md` endpoint will serve unrendered JSX. A build-time check that warns on JSX-in-MDX for files under `pages/patterns/_patterns/` would prevent this.
- **`AgentView` shows word count, not token count.** Words are fine but agents budget in tokens. Even a rough `words * 1.3` estimate would be more useful — bonus if you display "fits in [n]-token context."
- **Title cosmetic.** `AgentView` displays the title as `/patterns.md` (with dot) but the actual URL is `/patterns/md` (no dot). Minor mismatch.
- **`⭐ Featured pattern`** in markdown output. Emoji works, but `[FEATURED]` is more parseable for agents and friendlier to screen readers.

---

Happy to draft a follow-up PR (or extend this one) to add the discovery layer + seed `/llms.txt`. The `.md` endpoints are the hard part — the discovery signals are short and mostly mechanical.

pages/patterns/index.tsx
```diff
@@ -1,223 +1,173 @@
...
+          event-driven systems they live in. Each pattern is built on Inngest
+          primitives (steps, events, throttling, schedules, channels) and the
+          guarantees they provide.
+        </p>
+
+        <div className="flex max-w-xl gap-14 border-t border-subtle pt-8">
+          <div className="flex flex-col gap-1.5">
+            <span className="font-mono text-4xl text-basis">{totalPatterns}</span>
+            <span className="font-mono text-[11px] uppercase tracking-widest text-muted">
+              Primitives
```
SterlingChin: Label mismatch: `{totalPatterns}` (line 140) is a count of patterns but is labeled **Primitives** here, and `{sections.length}` is labeled **Categories** below. The earlier preview rendered "8 Patterns / 5 Primitives / 3 Languages" — current code renders "N Primitives / N Categories / 3 Languages."

Suggested fix: relabel `{totalPatterns}` as "Patterns", and either hardcode "5" for Primitives or drop that stat.

shared/Patterns/Featured.tsx
```diff
@@ -0,0 +1,106 @@
...
+              <svg width="14" height="14" viewBox="0 0 14 14" fill="none">
+                <path
+                  d="M3 7 L11 7 M7 3 L11 7 L7 11"
+                  stroke="currentColor"
+                  strokeWidth="1.4"
+                />
+              </svg>
+            </Link>
+            <Link
+              href={`/patterns#${section.id}`}
```
SterlingChin: This links to `/patterns#${section.id}`, but the index never renders section anchors (no `id="flow"` etc. on the page). Dead link as of this branch. Either render the section listings on the index, or change the destination to the section's first pattern.

shared/Patterns/patternsData.ts
```diff
@@ -0,0 +1,116 @@
...
+    hex: string;       // Raw hex for CSS custom properties
+    rgb: string;       // "R G B" for rgba mixing
+  };
+}
+
+export interface PatternSection extends PatternSectionMeta {
+  patterns: PatternItem[];
+}
+
+const PATTERN_SECTIONS: PatternSectionMeta[] = [
```
SterlingChin, line 33: 5 sections defined here (durable, flow, events, schedule, jobs), and all 5 currently have patterns assigned via MDX frontmatter. PR description says "Page structure reorganized into three sections." Either trim this file to 3 if Scheduling and Background Jobs should go away, or update the PR body to match.

pages/patterns/_patterns/flash-sales-and-bursty-workflows.mdx
```diff
@@ -0,0 +1,290 @@
...
+    });
+  }
+);
+```
+
+**Adapt it:**
+
+- Change `limit` and `period` to match your API's published rate limit. Leave 10-20% headroom.
+- Set `key` to whatever dimension your rate limit applies to (API key, account, carrier, region). Omit `key` if the limit is global.
+- Runs over the limit are enqueued and execute when capacity is available.
```
SterlingChin: Readers will immediately ask: how big can the throttle queue get, what's the ordering guarantee, and do queued runs ever expire? One sentence on queueing semantics in this Adapt-it block would head off the next question.

pages/patterns/_patterns/flash-sales-and-bursty-workflows.mdx
```diff
@@ -0,0 +1,290 @@
...
+      period: "30s",                 // wait 30s after the last event
+    },
+  },
+  async ({ event, step }) => {
+    // event contains the *last* event after the debounce window closes,
+    // so we're always working with the most recent data.
+
+    // Fetch the current order state, not the potentially stale event payload.
+    const order = await step.run("fetch-latest-order", async () => {
+      return await shopify.orders.get(event.data.order_id);
```
SterlingChin, line 161: The comment above (lines 156-157) says "we're always working with the most recent data" referring to the event payload — but here you wisely refetch via `shopify.orders.get(...)`. Worth one line explaining *why* you refetch (the payload from the last debounced event can still be stale by the time the function fires) — otherwise readers see the contradiction and aren't sure which to trust.

pages/patterns/_patterns/flash-sales-and-bursty-workflows.mdx
```diff
@@ -0,0 +1,290 @@
...
+  {
+    id: "capture-payment",
+    triggers: [{ event: "payment/authorized" }],
+    idempotency: "event.data.payment_id",   // one execution per payment, guaranteed
+  },
+  async ({ event, step }) => {
+    // Capture the charge. Idempotency ensures this only runs once per payment_id.
+    const capture = await step.run("capture-charge", async () => {
+      return await stripe.paymentIntents.capture(
+        event.data.payment_intent_id
```
SterlingChin, line 205: The idempotency expression two lines up uses `event.data.payment_id`, but this line passes `event.data.payment_intent_id` to Stripe. Two different identifiers in 6 lines without explanation — readers will pause here. Either reconcile (use one) or add a line clarifying why they differ (e.g. `payment_id` is your internal ID, `payment_intent_id` is Stripe's).

pages/patterns/_patterns/flash-sales-and-bursty-workflows.mdx
```diff
@@ -0,0 +1,290 @@
...
+```typescript
+import { inngest } from "./client";
+
+export const fulfillOrder = inngest.createFunction(
+  {
+    id: "fulfill-order",
+    triggers: [{ event: "order/fulfillment.ready" }],
+
+    // Don't process the same webhook twice
+    idempotency: "event.data.idempotency_key",
```
SterlingChin, line 254: The standalone idempotency example above uses `event.data.payment_id` (a natural key from the payload). Here the composed example introduces `event.data.idempotency_key`, a field that wasn't mentioned anywhere prior. One inline sentence — e.g., "set this at the event source, often a webhook delivery ID" — closes the loop.

pages/patterns/_patterns/reliable-scheduling-systems.mdx
```diff
@@ -1,8 +1,15 @@
+---
+title: Reliable scheduling systems
+subtitle: Fan out to hundreds or thousands of independently-retried jobs from a single trigger.
+pattern: events
```
SterlingChin, line 4: Title is "Reliable scheduling systems" but `pattern: events` (Event Coordination). The subtitle clarifies this is really fan-out, but "scheduling" in the title will read as a Scheduling pattern misfiled into Event Coordination. Consider renaming to something like "Reliable fan-out to many jobs."

shared/Docs/navigationStructure.ts
```diff
@@ -877,8 +878,24 @@ const sectionLearn: (NavGroup | NavLink)[] = [
         href: `/docs/local-development`,
       },
       {
-        title: "Patterns",
+        title: "Events and Triggers",
```
SterlingChin, line 881: The renamed "Events and Triggers" group here drops the link to `/docs/features/events-triggers/event-format` (Event payload format) that existed in the previous version of this group. If that page still exists, it's now orphaned from the sidebar. Was the removal intentional?

SterlingChin:
**[`pages/_app.tsx`]** — heads up: `useFirstTouch` was removed from this file in this PR (the `import` from `../shared/firstTouch` plus the `useFirstTouch()` call inside `MyApp`). That's the first-touch attribution hook (utm_*, ref, landing_url, referrer). Was this intentional? Removing it will cause site-wide attribution data loss, and it's unrelated to the patterns refresh — looks like a stale-base artifact.

(Posting as a general comment since GitHub wouldn't resolve the inline anchor on the context line near the deletion.)

SterlingChin:
**[`next.config.mjs`]** — `images.qualities: [75, 95]` was removed in this PR (was on line 472 of main, in the `images:` block right above `remotePatterns`). In Next 15 you get warnings (and on some configs broken optimization) when a component uses a `quality` prop value that isn't listed here. Was this intentional? Doesn't appear related to the patterns refresh — possibly a stale-base artifact.

Same file also removed the permanent redirect for `/2026-durable-execution-report` → `/content/ai-in-production-report-2026` in the `redirects()` block — flagging in case that was also accidental.

(Posting as a general comment since GitHub wouldn't resolve inline anchors on the context lines near the deletions.)

pages/patterns/index.tsx
Benanna2019: yes, I mentioned this many times in our convos. The others are not ready and I am trying to get one out at a time. The others don't need to be launched today and we are unlinking the 'hub page' for patterns to redirect to the only pattern we have right now which will change when others are launched.

PR #1564, opened 2026-05-01
Fix blog date display for YYYY-MM-DD frontmatter

LCraigie:
Not sure if we set this up this way intentionally, or if this is just a me<>Cursor problem, but the date for blogs was always the day prior. Cursor says: ISO date-only strings parse as UTC midnight; toLocaleDateString in US timezones showed the previous calendar day. Format date-only values in UTC so published dates match frontmatter. Applies to blog posts, blog metadata loader, changelog long-form dates, and updated-at lines.


djfarrelly (approved):
This is a weird thing that looks different locally vs prod, so this is a good change!

PR #1566, opened 2026-05-01
fix: redirect content download form to PDF after submit - AI in Prod Report LP

patrickholcomb9:
Updates the `redirectTo` URL for the `ai-in-production-report-2026` content asset so that users are sent directly to the PDF after submitting the download form.

## Changes
- `shared/contentAssets.ts`: replaced the placeholder blog post redirect with the PDF URL (`/assets/reports/2026-benchmark/2026-durable-execution-benchmark-report.pdf`)


whatupjohnb (approved):
Looks good to me

PR #1570, opened 2026-05-05
Fix logos mobile

whatupjohnb:
Logos were squished on mobile. Changed itso they proportionally scale now. Hid reminly.


LCraigie (approved):
lgtm ty!

PR #1596, opened 2026-05-14
AI in Prod: improve report landing fold visibility

patrickholcomb9:
Changes to AI in Prod landing page to make it more clear to users that there is more info/key findings below (previously was hard to tell that there more info below fold).

- Compact hero panel — reduced padding, logo sizes, Inngest wordmark size, heading size, copy gap, and removed min-h-[85vh] so the hero takes only as much space as its content needs. The Key Findings section now peeks just above the fold on most viewports, naturally signaling there's more to scroll.
- Fix scroll indicator bug — the bounce arrow was positioned at bottom-8 (32px) but the sticky footer bar is 73px tall, so the arrow was completely hidden behind it. Changed to bottom-24 (96px) so it's always visible.
- Reduce Key Findings top padding — pt-20 md:pt-28 → pt-8 md:pt-10 so the section header appears immediately when users scroll rather than after a large gap of empty space.


patrickholcomb9:
Superseded by https://github.com/inngest/website/pull/1597

PR #1611, opened 2026-05-27
DEV-337: Add Vercel integration troubleshooting page

Benanna2019:
Trying to have a good doc in place for users as this is a common customer support question

### DIFF
--- a/pages/docs/deploy/vercel-troubleshooting.mdx
+++ b/pages/docs/deploy/vercel-troubleshooting.mdx
@@ -1,0 +1,87 @@
+import { Callout, CodeGroup } from "src/shared/Docs/mdx";
+
+export const description = "Troubleshoot common Vercel integration issues: preview branches, key rotation, Marketplace vs direct accounts, multi-project setup, and WAF interactions."
+
+# Vercel integration troubleshooting
+
+This page covers the most common issues when running Inngest on Vercel. If your functions aren't syncing, your preview environments aren't working, or you're seeing unexpected behavior after changing keys, start here.
+
+---
+
+## Preview branches not created after key rotation
+
+After rotating your signing key in the Inngest dashboard, preview branch environments may stop syncing. New deploys succeed on Vercel but Inngest doesn't pick up the updated functions.
+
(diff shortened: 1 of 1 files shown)


pages/docs/deploy/vercel-troubleshooting.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+## Vercel Marketplace vs direct account
+
+There are two ways to connect Inngest to Vercel:
+
+**Vercel Marketplace install** creates your Inngest account through Vercel's integration marketplace. Billing goes through Vercel. Environment variables are set automatically.
+
+**Direct Inngest account** is created at app.inngest.com. You install the Vercel integration separately. Billing goes through Inngest directly.
+
+If you started with a Marketplace install and want to switch to a direct account (or vice versa), contact support. The two account types use different authentication flows and mixing them causes sync failures.
```
campbell88, line 29: there is no way for us to migrate marketplace users to direct account users (although I wish there was). So i think the copy would be more like you would need to start from scratch and follow our guide on deploying to vercel, like using the Vercel integration from the integrations page in our dashboard

also i feel like we could benefit from having a section about the difference between marketplace accounts and direct accounts. like list the pros and cons. We cannot discourage them not to use marketplace as that may affect our agreement w Vercel, so we can think of this at a later time

pages/docs/deploy/vercel-troubleshooting.mdx
```diff
@@ -0,0 +1,87 @@
...
+3. Check the [Inngest dashboard Apps page](https://app.inngest.com) for sync errors. Failed syncs show an error message explaining what went wrong.
+4. If using Deployment Protection, make sure you've configured the [protection bypass](/docs/deploy/vercel#bypassing-deployment-protection).
+
+---
+
+## Runs timing out on Vercel
+
+Vercel serverless functions have execution time limits that depend on your plan. If your function runs are timing out:
+
+1. Set `maxDuration` on your serve endpoint to match your Vercel plan's limit (up to 300 seconds on Pro).
```
campbell88, line 66: maybe a link to vercel's limit is better than hard coding # of seconds, this has bit us in the butt before

PR #1618, opened 2026-06-01
Sessions docs

jakobevangelista:
docs for sessions https://github.com/inngest/inngest-js/pull/1547

preview: https://website-git-jakob-sessions-docs-inngest.vercel.app/docs/features/events-triggers/sessions


pages/docs/features/events-triggers/sessions.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+```ts
+await inngest.send({
+  name: "app/agent.step.completed",
+  data: {
+    stepId: "step_1",
+  },
+  sessions: {
+    conversation_id: "conv_1234",
+    attempt: 2,
```
djfarrelly: "attempt" reads more like metadata and not a session id, perhaps we could rename this
```suggestion
    thread_id: 29563,
```
Or if I'm mis understanding the feature, when would someone use a boolean for this?

jakobevangelista: great point, got me thinking when would someone use a boolean? basically never

`attempt/human_reviewed` are low cardinality metadata, not session ids, and we're treating that as misuse now (thread [here](https://inngest.slack.com/archives/C0AFJB68X99/p1781189607738199))
 
i'll swap it for a real id like your `thread_id`. it was just there to show numbers work, which they still do
 
also got me thinking i should remove supporting `boolean` as a type because that is inherently low cardinality

pages/docs/features/events-triggers/sessions.mdx
```diff
@@ -0,0 +1,185 @@
+import { Callout } from "src/shared/Docs/mdx";
+
+export const metaTitle = "Inngest Sessions: Group Related Function Runs";
+export const description = "Group related function runs with sessions so they are easier to find and inspect in the dashboard.";
+
+# Sessions
```
djfarrelly, line 7: @jakobevangelista we should add a label or callout that sessions were added in TS SDK v4.7

pages/docs/features/events-triggers/sessions.mdx
```diff
@@ -0,0 +1,185 @@
...
+
+The object key is the session key. The value is the session ID.
+
+In the example above:
+
+- `conversation_id` is the session key
+- `conv_1234` is the session ID
+
+You can use the same session key across many events. Each unique session ID becomes a session in the dashboard.
+
```
djfarrelly, line 60: @jakobevangelista  We should include some sort of callout either here or below mentioning that sessions can be added to webhook transforms as well - this is something my coding agent missed when I used this page. It doesn't need to be a full section yet, but just mentioning it and then linking to the webhooks docs if the dev wants more info on transforms, that works well.

pages/docs/features/events-triggers/sessions.mdx
jakobevangelista: https://github.com/inngest/website/pull/1618/commits/1fa9ddc32ffcb1fe499593212c1c35ab794f6042

pages/docs/features/events-triggers/sessions.mdx
jakobevangelista: https://github.com/inngest/website/pull/1618/commits/b50a8433d4c9979cf0c0d94e3d31859bb08b05db

anafilipadealmeida (approved):

PR #1624, opened 2026-06-03
Add Branch vs Custom Environments concept page

Benanna2019:
## Summary

Concept page explaining the difference between Branch Environments and Custom Environments. Confusing the two is one of the most common setup issues from support.

Covers: what each type is, the common mistake (preview deployments targeting a custom env instead of branch envs), when to use which.

Source: Dan's support inbox scan (T-7550). Support team's reply was near-verbatim the article.

Resolves DEV-338

### DIFF
--- a/pages/docs/platform/branch-vs-custom-environments.mdx
+++ b/pages/docs/platform/branch-vs-custom-environments.mdx
@@ -1,0 +1,61 @@
+import { Callout } from "src/shared/Docs/mdx";
+
+export const description = "Understand the difference between Branch Environments and Custom Environments in Inngest."
+
+# Branch vs Custom Environments
+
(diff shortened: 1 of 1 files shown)


pages/docs/platform/branch-vs-custom-environments.mdx
```diff
@@ -0,0 +1,53 @@
+import { Callout } from "src/shared/Docs/mdx";
+
+export const description = "Understand the difference between Branch Environments and Custom Environments in Inngest."
+
+# Branch Environments vs Custom Environments
```
djfarrelly, line 5: @Benanna2019 Where is this linked to from? How does this relate to the main environments guide? I'm not sure how these two pages should best relate as not to duplicate content - or if this should fit more succinctly into a the "FAQ" page: https://www.inngest.com/docs/faq

PR #1621, opened 2026-06-03
SEO: add site-wide canonical tags and block ref/UTM params from crawl

patrickholcomb9:
## Changes

**`middleware.ts`**
- Extended matcher from blog/docs-only to all routes
- Forwards `x-pathname` request header (clean path, no query params) to layout
- Emits `Link: <canonical>; rel="canonical"` HTTP response header on all pass-through responses

**`app/layout.tsx`**
- Added `CanonicalTag` async server component that reads `x-pathname` from `headers()` and emits `<link rel="canonical" href="https://www.inngest.com/...">` in `<head>` on every page
- Added `metadataBase` to root metadata export

**`next-sitemap.config.js`**
- Added `robotsTxtOptions.policies` disallowing `?ref=` and `?redirect_url=` param crawling for all crawlers
- Preserved existing AI crawler allow/disallow rules (GPTBot, ClaudeBot, etc.)

PR #1627, opened 2026-06-03
Send canonical URL header for markdown pages

djfarrelly:
Build on #1621 to ensure that all markdown pages, which are duplictes have canonicals.


patrickholcomb9:
this all looks good to me, thank you for walking through these updates w me Dan!

PR #1628, opened 2026-06-03
Optimize image for resource popover

djfarrelly:
Previously, the popover was loading a 1MB image. This uses the image optimization URLs


patrickholcomb9 (approved):
sweet ty for flagging & the quick turnaround on this! lgtm

PR #1629, opened 2026-06-04
perf: Minor optimizations for blog pages

djfarrelly:
We're aiming at improving LCP on blog post pages


patrickholcomb9 (approved):
all cleanup & performance updates look good to me! ty dan!

PR #1632, opened 2026-06-05
Publish 2026 AI in Production benchmark report as a "blog."

LCraigie:
Adds PDF-aligned report layout, section breaks, charts, and TOC so the survey findings are readable on the site without leaving the blog.

Itll start as a 'blog' but I think we can re-label it as a report in the resources center later on.

### DIFF
--- a/app/content/ai-in-production-report-2026/graphs.ts
+++ b/app/content/ai-in-production-report-2026/graphs.ts
@@ -5,6 +5,10 @@ export type ReportGraph = {
   shareText: string;
 };
 
+/** Native dimensions of OG share-card chart PNGs — used for next/image. */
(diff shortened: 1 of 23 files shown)


patrickholcomb9 (approved):
lgtm!

Updates from PH:

- Background texture quality reduced for pagespeed/performance; colors are a litttle less vibrant now, but imo it still looks fine & onbrand? if we think it's too much of a change i can find some middle ground on texture quality level
- confirmed that global banner + blog CTAs (at bottom of blogs) now point to the /blog report (not the gated LP)
- added noindex & canonical tag to /content/ (gated report), so that LP wont be crawlable by search engines, users can only find it if we direct them there, and Google will know that if it crawls the gated report, it should ignore that page and use the /blog report as the "true" URL

PR #1633, opened 2026-06-05
Deferred Functions: reference docs, changelog, and nav entry

Benanna2019:
## Summary

Adds Deferred Functions documentation, extracted from Linell's work on PR #1603 into a clean branch against current main.

Three changes:

- **Deferred Functions reference** (`/docs/reference/typescript/v4/functions/deferred-functions.mdx`): Full API reference for `createDefer` and `defer()`. Covers defining deferred functions, calling them, schemas, error handling, and sharing across parent functions. Diataxis: Reference.
- **Changelog entry** (`content/changelog/2026-06-05-deferred-functions.mdx`): Announces the beta release.
- **Nav entry**: Added under TypeScript SDK v4 in the reference sidebar, tagged as beta.

## Context

Linell originally merged this into the `ben/agent-evals-scoring-docs` branch (PR #1603). That branch had diverged significantly from main. This PR applies the same content cleanly against current main with the updated nav structure (patterns, etc.).

Co-authored-by: Linell Bonnette <tlbonnette@gmail.com>

PR #1639, opened 2026-06-09
Add deferred work concept page and how-to guide under Agents

Benanna2019:
## Summary

Two new docs pages for deferred functions under the Agents section in Learn.

**Deferred work** (`/docs/ai-patterns/deferred-work`) — Diataxis Explanation. Covers the mental model ("because of" vs "as part of"), how deferred runs relate to the parent, and where defer fits alongside scoring, experiments, and temporal signals. Includes comparison table against alternative approaches.

**Defer AI side effects** (`/docs/ai-patterns/defer-ai-side-effects`) — Diataxis How-to. Five steps: define a deferred function, call defer from your agent, register in serve, add more deferred functions, verify in the dashboard. Practical walkthrough with LLM-as-judge scoring and cost tracking examples.

Both added to the Agents nav group with beta tags. Both cross-link to the Deferred Functions reference (PR #1633).

Resolves DEV-383


SterlingChin (changes requested):
Reviewed ahead of the handoff — content is in great shape, three blockers before merge:

**1. Duplicate page still in the PR**
`pages/docs/ai-patterns/deferred-work.mdx` is orphaned (no nav references) but would ship as a live URL with ~80% duplicate content of `guides/deferred-functions.mdx` (same mental model section, same comparison table). Should be deleted from the branch.

**2. Concept page contradicts the merged reference**
`guides/deferred-functions.mdx` ("When to use defer" → Cancellable work) says `defer()` returns a handle with `.cancel()`. The merged reference (`/docs/reference/typescript/v4/functions/deferred-functions`) says `defer(...)` is synchronous, fire-and-forget, and returns `void` — and the how-to in this PR says the same. The cancel bullet documents an API that doesn't exist; looks like a leftover from an earlier draft.

**3. Merge-order dependency on #1634**
`guides/deferred-functions.mdx` links to `/docs/patterns/durable/deferred-cleanup-and-rollbacks`, which only exists in #1634. Merge #1634 first or this 404s.

Non-blocking:

- `navigationStructure.ts` has ~1,084 deleted lines for what should be a 2-link change. I verified it's functionally equivalent (248→250 hrefs, nothing dropped), but it strips every comment including the one explaining why `tsRef` links concrete versioned routes (the Vercel `_next/data` rewrite issue). That comment prevents a future regression — the nav diff should be minimized back to just the additions.
- `defer-ai-side-effects.mdx` hardcodes GPT-4o pricing ($0.005/1k, same rate for prompt+completion). Fine as illustrative code but worth a comment marking it illustrative so it doesn't age into being confidently wrong.

The writing itself is solid — the "because of vs. as part of" mental model is crisp and the how-to progression is clean. Once these are addressed this is good to go.

SterlingChin:
I'll push an update to this branch addressing the review items directly (per handoff with @Benanna2019 before his leave) — keeping the #1634 merge-order note for whoever lands these.

PR #1643, opened 2026-06-09
CLI docs: reference, debug guide, and coding agent guide

Benanna2019:
## Summary

Three new docs pages for the CLI + Agentic API launch (June 9-11).

**CLI reference** (`/docs/cli`) — Diataxis Reference. Quick start, authentication (env-scoped API keys), dev server vs cloud targeting, available commands table, output format. Based on the Agentic API PRD and Jacob's shipped CLI examples.

**Debug a function run from your terminal** (`/docs/guides/debug-with-cli`) — Diataxis How-to. Four steps: get the run summary, fetch the trace, pipe to jq, test a fix locally. Uses realistic JSON output showing a failed step.

**Use the CLI with coding agents** (`/docs/ai-patterns/cli-for-coding-agents`) — Diataxis How-to. Four steps: set the API key, add Inngest context to your agent (CLAUDE.md / .cursorrules snippet), debug a failing function, test the fix. The AX angle from the PRD.

All three cross-link to each other, to the REST API docs at api-docs.inngest.com, and to relevant existing pages (traces, error handling, API keys).

Nav entries not included in this PR to avoid conflicts with other in-flight nav changes. Will add in a follow-up.

Resolves DEV-385

### DIFF
--- a/pages/docs/ai-patterns/cli-for-coding-agents.mdx
+++ b/pages/docs/ai-patterns/cli-for-coding-agents.mdx
@@ -33,10 +33,10 @@ Tell the agent how to use the CLI. Add this to your project's `CLAUDE.md`, `.cur
 
 Use the Inngest CLI to inspect function runs and traces:
 
-- Get a run: `npx inngest-cli@latest api --prod get-run <run_id>`
+- Get a run: `npx inngest-cli@latest api --prod get-function-run <run_id>`
 - Get a trace: `npx inngest-cli@latest api --prod get-function-trace <run_id> --include-output`
 - Get runs from an event: `npx inngest-cli@latest api --prod get-event-runs <event_id> --include-output --limit 5`
-- Invoke locally: `npx inngest-cli@latest api invoke-function-by-slug <function_slug> --data '{...}'`
+- Invoke locally: `npx inngest-cli@latest api invoke-function <app_id> <function_id> --data '{...}'`
 
 Output is JSON. Use jq to filter.
 ```
@@ -66,7 +66,7 @@ It reads the trace, identifies the failing step, sees the error output, and prop
 After the agent makes a code change, it can invoke the function locally against the dev server:
 
 ```bash
-npx inngest-cli@latest api invoke-function-by-slug my-app-process-order \
+npx inngest-cli@latest api invoke-function my-app process-order \
   --data '{"orderId": "test-123"}'
 ```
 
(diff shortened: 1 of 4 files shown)


SterlingChin (approved):
Approving to keep launch unblocked — the structure and narrative here are exactly right, especially the debug guide. One thing needs confirmation before this merges, because it decides whether readers' copy-pastes work on day one.

## Confirm the command surface

The docs use the Agentic API PRD's surface, but the shipped CLI (1.27.0, current `@latest`) generates its commands from the v2 OpenAPI operationIds and exposes different names:

| In these docs | Shipped CLI 1.27.0 |
|---|---|
| `api get-run <run_id>` | `api get-function-run <run_id>` |
| `api invoke-function-by-slug <slug>` | `api invoke-function <app_id> <function_id>` (no slug-based invoke exists) |
| `--api-url` / `INNGEST_API_URL` | `--api-host` / `--api-port` |

I checked `proto/api/v2/service.proto` across all branches of the inngest repo: there's no `InvokeFunctionBySlug` RPC anywhere, and the pending v1.28.0 release doesn't touch the apiv2cli command names. So as far as I can tell, the PRD's slug invoke and friendlier names haven't been built yet.

**Ask:** @Benanna2019 / Jacob — can you confirm whether the PRD CLI surface (slug invoke, `get-run`, `--api-url`) is landing in the launch release? If yes, merge as-is once it's released. If not, these should be renamed to the shipped commands before merge — happy to push that commit, it's mechanical and I've verified every replacement against the live CLI. Note the launch blog uses the same names, so whichever way this goes, the blog needs to match.

## Smaller items (non-blocking, fix whenever)

- The run-summary JSON sample shows `function.slug` — the shipped v2 response is `function: {id, name}` (slug is in the PRD shape but not the published spec).
- Step 4 in both guides verifies with `get-event-runs <event_id>` where the event ID comes from nowhere — the invoke response returns a `runId`, so `get-function-run <runId>` is the natural follow-up.
- The `navigationStructure.ts` reformat strips all comments, including the `tsRef` one explaining the Vercel `_next/data` rewrite gotcha — worth restoring. (I diffed the href sets between main and this branch: nothing lost, 161 links plus the 3 new ones, so the reformat is functionally safe.)
- Commands table covers 4 of the 18 live subcommands — a one-liner noting "every v2 endpoint is a subcommand" plus a mention of `sync-app` and the Insights commands would round out the agent story.

PR #1648, opened 2026-06-09
Pattern: Experiment, score, and compare AI models

Benanna2019:
## Summary

Comprehensive AI experimentation pattern in the style of the flash-sales-and-bursty-workflows pattern. Shows three AI primitives from the Inngest SDK composing together in one flow:

- **`group.experiment()`** splits traffic between GPT-4o and Claude Sonnet (50/50)
- **`defer()`** kicks off a cost tracker that calculates per-variant, per-customer token spend after the response ships
- **Insights** aggregates run data so you can query latency, failure rate, and cost across variants

The function serves a response immediately. The cost tracking runs in the background as a deferred function with independent retries. Data accumulates over days or weeks. You make a decision based on real production traffic.

Includes the full function (agent + deferred cost tracker), an explanation of how the pieces compose, and four adaptation sections: compare prompts instead of models, compare RAG strategies, gradual model rollout, and deferring more than just costs.

No scoring API dependency. Everything in this pattern ships in the next two weeks (experiments + defer + insights).

Diataxis: Explanation (pattern). Matches the structure and comprehensiveness of the flash sales pattern.

### DIFF
--- a/shared/Patterns/_patterns/experiment-score-and-compare-ai-models.mdx
+++ b/shared/Patterns/_patterns/experiment-score-and-compare-ai-models.mdx
@@ -1,21 +1,20 @@
 ---
-title: Experiment, score, and compare AI models
-subtitle: Use experiments to split traffic between models, defer scoring to run after the response, and query results to pick the winner.
+title: Experiment and compare AI models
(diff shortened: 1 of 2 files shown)


Benanna2019:
Since Sterling handled the experiment side of things I will close this and pull score docs into another PR

PR #1672, opened 2026-06-11
fix(nav): responsive promo card in Resources dropdown

patrickholcomb9:
The featured promo card in the Resources nav dropdown had a fixed `w-[400px] shrink-0`, causing the nav columns to get squeezed at smaller desktop viewport widths.

## Changes
- Removed `shrink-0` from the promo card and its Link wrapper, allowing the card to shrink when space is constrained
- Added `min-w-[240px]` as a floor so the card doesn't collapse too small
- Updated the `sizes` prop on the promo image so Next.js serves the appropriate resolution as the card scales down

### DIFF
--- a/components/v1/NavMenuPanel.tsx
+++ b/components/v1/NavMenuPanel.tsx
@@ -49,7 +49,7 @@ export default function NavMenuPanel({
         "relative isolate w-full overflow-clip rounded-[6px] bg-carbon-1000",
         PANEL_BORDER,
         hasPromo
-          ? "flex gap-6 xl:gap-14 px-6 xl:px-8 py-8"
+          ? "flex flex-col xl:flex-row gap-6 xl:gap-14 px-6 xl:px-8 py-8"
           : compact
             ? "px-8 py-5"
             : "px-8 pt-8 pb-10",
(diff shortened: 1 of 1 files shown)


patrickholcomb9:
The featured promo card in nav dropdowns had a fixed `w-[400px] shrink-0`, causing text columns to get squeezed at smaller desktop viewport widths.

## What we tried
- Removing `shrink-0` to allow the card to shrink — card shrank but columns shrank alongside it, causing text wrapping
- Using `w-[min(400px,38%)]` (CSS min() in Tailwind arbitrary value) — generated invalid CSS, card collapsed to content size (~14% due to double-compounding when both the Link wrapper and card div had the percentage width)
- Using `w-[38%] max-w-[400px] flex-none` — fixed the double-compounding but `flex-none` prevented shrinking on narrower Platform/Use Cases panels, causing the card to overflow and get clipped
- Adding `shrink-0` to columns/items and `shrink` to the card — card compressed correctly on narrow panels, but at ~870px the 400px card still left only ~340px for 3 text columns, not enough to prevent wrapping
- Removing Feat Images from smaller viewports

## Final approach
- Hide the promo card entirely below `xl` (1280px) — columns get full panel width at smaller desktop sizes
- At `xl`+, card renders at `w-[400px]` with `shrink` so it can compress if the panel is tight
- `MenuItemList` and `MenuColumns` are `shrink-0` so only the card absorbs flex overflow
- Gap and padding tighten below `xl` (`gap-6 px-6`) and expand at `xl`+ (`gap-14 px-8`)
- Updates `sizes` prop on the promo image so Next.js serves the right resolution as the card scales

LCraigie (approved):

PR #1676, opened 2026-06-11
feat(seo): add X-Robots-Tag noindex headers to LLM markdown routes

patrickholcomb9:
## What
Adds `X-Robots-Tag` response headers to the `/docs-markdown/` and `/blog-markdown/` route handlers to block Googlebot and Bingbot from indexing these pages.

## Why
These routes serve lightweight markdown mirrors of our docs/blog intended for AI crawlers. Without noindex headers, Google treats them as near-duplicate, low-styling versions of the main HTML pages and drops them into "Crawled - currently not indexed", wasting crawl budget.

## Notes
- Uses `Headers.append()` to set multiple `X-Robots-Tag` headers (one per bot), since a plain object cannot have duplicate keys
- Both routes already set a `Link: <canonical>` header pointing to the real page — this adds the noindex layer on top
- ⚠️ A follow-up PR (`pat/seo-robots-txt-llm-block`) adds `Disallow` rules in `robots.txt` for these directories — that PR should only be merged ~2 weeks after this one ships, to give Googlebot time to re-crawl and register the noindex before being blocked

### DIFF
--- a/app/(llms)/blog-markdown/[slug]/route.ts
+++ b/app/(llms)/blog-markdown/[slug]/route.ts
@@ -99,13 +99,17 @@ export async function GET(
     const markdownBody = await convertMdxToMarkdown(content);
     const output = `${frontmatterLines.join("\n")}\n\n${markdownBody}`;
 
-    return new Response(output, {
-      headers: {
-        "Content-Type": "text/markdown;charset=UTF-8",
-        "Cache-Control": "s-maxage=3600, stale-while-revalidate",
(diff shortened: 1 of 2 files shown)


djfarrelly (approved):
This is good to merge @patrickholcomb9 - it looks like the robots tags are being merged correctly, but we're seeing duplicate headers for Link - this shouldn't be an issue, but with this approach, we may want to remove the logic from the `middleware` in the future to simplify the code overall. 

Also, aspirationally, it would be ideal to have zero middleware on the website which can make things faster to load, but it's a small impact right now - not a worry of ours right now at all.

Image

PR #1680, opened 2026-06-12
fix: convert v4 docs reference redirects from 302 → 301 and collapse two-hop chains

patrickholcomb9:
## What

Fixes SEO-impacting temporary redirects on all TypeScript SDK reference paths, 
including those added in the Feb 27 v4 docs migration.

## Why

302 (temporary) redirects don't reliably pass link equity — Google may continue 
indexing old URLs instead of the canonical v4 pages. All of these paths are 
permanent migrations and should be 301s.

## Changes

- **Removed** the duplicate `permanent: false` block from the `redirects()` function 
  body — these 23 entries were shadowing the `permanent: true` versions in 
  `permanentRedirects`, so the 302s were always winning
- **Collapsed two-hop chains**: old paths were 302ing to versionless 
  `/docs/reference/typescript/*` URLs, which were then internally rewritten to 
  `/docs/reference/typescript/v4/*`. Destinations now point directly to v4 in one hop
- **Fixed** `/docs/reference/typescript/v4` destination from `/client/create` → `/intro`
- **Added** `/docs/reference/typescript` and `/docs/reference/typescript/v3` as 
  explicit 301s (were 302-only before, no permanent counterpart existed)

### DIFF
--- a/components/v1/sections/BackgroundJobs/Reliability.tsx
+++ b/components/v1/sections/BackgroundJobs/Reliability.tsx
@@ -94,7 +94,7 @@ const FEATURES: Feature[] = [
     id: "typescript",
     title: "TypeScript-native",
     body: "Define event payload types once. End-to-end type safety from the event sender all the way into your function body.",
-    href: "https://www.inngest.com/docs/guides/singleton",
+    href: "https://www.inngest.com/docs/reference/typescript/v4/intro",
   },
 ];
 
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -105,75 +105,74 @@ const permanentRedirects = [
     "/docs/reference/typescript/functions/metadata",
   ],
 
-  // TypeScript SDK versioned docs - landing page redirect
-  [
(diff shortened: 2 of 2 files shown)


SterlingChin (approved):
Ship it

SterlingChin (approved):

SterlingChin:
Closing this as superseded by #1683, which shipped the 302 -> 301 docs redirect fix and collapsed the v4 reference redirect chains.

The only useful remaining change from this branch is the one-line Background Jobs card link update in `components/v1/sections/BackgroundJobs/Reliability.tsx`. I’m splitting that into a fresh, focused PR and will link back here so the provenance from Pat’s branch is preserved.

SterlingChin:
Follow-up PR for the one remaining non-redirect change from this branch is here: #1704.

#1683 already shipped the 302 -> 301 redirect fix, so #1704 keeps only Pat’s Background Jobs card link cleanup in a focused one-file PR.

PR #1683, opened 2026-06-15
fix: make docs migration redirects permanent

SterlingChin:
## Summary

This PR ships the docs redirect SEO fix from Pat’s analysis: the v4 docs migration/reference redirects were using `permanent: false`, which made them temporary 302 redirects. That was causing SEO issues and, in some cases, redirect chains through the versionless TypeScript docs rewrite.

This updates the affected docs migration redirects to permanent 301s and collapses legacy short paths directly to their final `/docs/reference/typescript/v4/...` destinations.

## Context

- Pat’s original work/branch: https://github.com/inngest/website/tree/pat/fix-302-to-301-redirects
- Notion project brief: https://app.notion.com/p/37db64753bbd80dabf8ad5760ebce791
- Linear ticket: https://linear.app/inngest/issue/DEV-429/ship-301-redirects-for-v4-docs-migration-paths
- Redirect review sheet: https://docs.google.com/spreadsheets/d/1Fr1gqDn63btmVLCkMzX3rdJm5lMAz52nCxEDnB9242E/edit?usp=sharing

## Changes

- Moves the v4 docs reference redirect set into `permanentRedirects` so they return 301s.
- Removes the duplicate temporary redirect block from `redirects()`.
- Points legacy reference URLs directly at final v4 docs paths, for example:
  - `/docs/reference/functions/create`
  - `/docs/reference/events/send`
  - `/docs/reference/serve`
- Keeps the v3 migration redirect pointing at its final v3 migration destination.

## Validation

- Ran a config-level redirect sanity check against `next.config.mjs`
  - verified 24 docs migration redirects are permanent
  - verified affected legacy paths are single-hop redirects
- Ran `pnpm exec prettier --check next.config.mjs`
- Ran `git diff --check`

## Follow-up

After deploy, we should submit the affected URLs in Google Search Console URL Inspection as part of DEV-429.

### DIFF
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -105,82 +105,87 @@ const permanentRedirects = [
     "/docs/reference/typescript/functions/metadata",
   ],
 
-  // TypeScript SDK versioned docs - landing page redirect
+  // TypeScript SDK versioned docs - landing page redirects (301 permanent)
+  ["/docs/reference/typescript", "/docs/reference/typescript/intro"],
+  ["/docs/reference/typescript/v4", "/docs/reference/typescript/v4/intro"],
+  ["/docs/reference/typescript/v3", "/docs/reference/typescript/v3/intro"],
+  // Legacy short paths - redirect directly to v4 TypeScript docs (collapsed from two-hop chain)
   [
-    "/docs/reference/typescript/v4",
+    "/docs/reference/client/create",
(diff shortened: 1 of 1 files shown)


LCraigie (approved):
rubber stamping because sterling said i can

djfarrelly (approved):

PR #1699, opened 2026-06-17
Instruct users not to use connection pooling for the Neon integration

scottnuma:
- We require connections without connection pooling. Surface this to the user in the instructions for the smoothest installation
- I inserted the warning block below. See the before at https://www.inngest.com/docs/features/events-triggers/neon
Screenshot 2026-06-17 at 11 15 17 AM


Benanna2019 (approved):
@scottnuma I believe this is good to go once main gets merged in. It looks like the issue breaking the build was handled in another PR

PR #1716, opened 2026-06-22
docs: document experimentRef, inngest.score.experiment(), and createScorer experiment scoring

AndyInternet:
Documents the new experiment-scoring surface across the steps/workflows guides and the v4 TypeScript reference.

group.experiment() now returns { result, variant, experimentRef }. Updated the experiments guide and reference to destructure it and explain the experimentRef handle (replaces the old withVariant examples).
inngest.score.experiment(): new section in the scoring guide and reference for scoring a variant from outside the callback, including the run-scope rule.
Deferred scorers: defer({ experiment }) routes a createScorer result to the experiment. New section in the deferred-scoring guide plus the experiment option on the defer() reference.
Touches running-experiments, scoring, and deferred-scoring guides, and the group-experiment + scoring reference pages. These are MDX, so the code blocks aren't snippet-validated.

SDK changes: https://github.com/inngest/inngest-js/pull/1597


Linell (approved):

SterlingChin:
Thanks for getting these docs wired up to the SDK change. I did a broader docs impact pass using inngest-js#1597 as the source of truth, and I think there are two follow-ups worth handling before merge:

1. The return-shape change should be reflected in every guide example, not only the new reference section. Since `group.experiment()` now always resolves to `{ result, variant, experimentRef }`, the examples in `running-experiments.mdx` that still do `const summary = await group.experiment(...)`, `const response = ...`, and `const answer = ...` read as if the selected variant result is returned directly. I would update those to destructure `result` where the surrounding prose means the model response, for example `const { result: summary } = await group.experiment(...)`.

2. The canonical `defer()` reference should also get the new option. The PR documents `experiment: experimentRef` under the scoring reference, but `pages/docs/reference/typescript/v4/functions/deferred-functions.mdx` still lists only `{ function, data }` for `defer(id, options)`. Because the defer blog/changelog link to that page as the full API reference, it should include `experiment?: ExperimentRef` and explain that it attributes a deferred scorer result to the selected experiment variant.

Small polish while in there: the `experimentRef` return-value text in `group-experiment.mdx` links `defer()` to the scoring page; that should probably point to `/docs/reference/typescript/v4/functions/deferred-functions`. The scoring reference meta title also still says `step.ai.score()`, which seems stale now that the page covers `inngest.score()`, `step.score()`, `createScorer()`, and `inngest.score.experiment()`.

CI note: the failing link check appears unrelated to this PR. It is the existing LinkedIn 404 in `content/blog/eliminating-latency-ai-workflows.mdx`.

Benanna2019 (approved):

PR #1742, opened 2026-06-26
Add "Score agents on real outcomes" pattern page

mitchellalderson:
New AI Evals pattern page covering scoring agents on real-world outcomes: inline scoring, deferred scoring against a runId, human feedback, session scoring, and reusable scorers via createScorer(). Registered in patternsData.ts under the ai-evals category.


Benanna2019 (dismissed):
This one needs to include (if it goes out before Ana's PR to ungate content around scoring and sessions) to add the ?unrealeased=score parameter. 

Also, I mentioned this in a comment on the more recent PR, but for the patterns going forward I know the team wants to have Inngest's observability story highlighted as a part of this "AI Evals and Observability" narrative. 

So having some screenshots of the dashboard of scoring and deferred scoring would be huge

LCraigie (approved):

PR #1747, opened 2026-06-26
Document scoreMiddleware

scottnuma:
- Users need to register scoreMiddleware before using step.score
- Let's document that more explicitly in our examples and function docs

### DIFF
--- a/pages/docs/features/inngest-functions/steps-workflows/scoring.mdx
+++ b/pages/docs/features/inngest-functions/steps-workflows/scoring.mdx
@@ -19,7 +19,10 @@ This is useful for anything where "did it work?" isn't a binary answer: AI accur
 
 ## Score the current run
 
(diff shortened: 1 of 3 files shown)


pages/docs/features/inngest-functions/steps-workflows/scoring.mdx
```diff
@@ -34,14 +37,49 @@ export default inngest.createFunction(
     });
 
     // Score this run based on whether guardrails passed
-    await step.score("score-guardrail-pass", { name: "guardrail-pass", value: passed ? 1 : 0 });
+    await inngest.score({ name: "guardrail-pass", value: passed ? 1 : 0 });
```
amh4r, line 40: What was wrong with the step? We want to encourage running client methods in steps in case something fails afterwards and there's a retry

PR #1750, opened 2026-06-29
Add Score agent accuracy pattern (gated behind ?unreleased=score)

mitchellalderson:
New AI Evals pattern explaining how to grade an agent against ground truth (exact match, Jaccard overlap, numeric tolerance) and record the score on the run that produced it.

Gated behind ?unreleased=score: hidden from the nav, landing, category pages, and the markdown/agent surfaces server-side, revealed client-side when the label is present. Adds an optional `unreleased` field to the pattern index and filters it through getPatternSections().

### DIFF
--- a/pages/api/patterns/[pattern]/md.ts
+++ b/pages/api/patterns/[pattern]/md.ts
@@ -6,6 +6,7 @@ type PatternFrontmatter = {
   title?: string;
   subtitle?: string;
   tags?: string[];
+  unreleased?: string;
 };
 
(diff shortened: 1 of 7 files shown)


Benanna2019 (approved):

Benanna2019:
@mitchellalderson Something I forgot to mention when reviewing this is that a big part of what we are wanting to push with the AI Evals patterns is the Inngest Observability story. So something like screenshots of scoring in the dashboard would speak to that.

PR #1801, opened 2026-07-22
Fix broken social preview image for blog posts with CDN image

mitchellalderson:
Blog OG/Twitter meta tags (and JSON-LD) were built by unconditionally prepending NEXT_PUBLIC_HOST to the frontmatter `image`. For posts using an absolute cdn.inngest.com URL this produced a malformed `https://www.inngest.comhttps://cdn.inngest.com/...` that X/LinkedIn reject. The on-page hero was unaffected since it uses next/image with the CDN host allowlisted.

Reuse the existing getFullURL helper (src/utils/social.ts), which passes absolute URLs through untouched and only absolutizes relative /assets paths. Also resolves the Vercel preview host so scrapers work on previews.


patrickholcomb9 (approved):
lgtm!

PR #1805, opened 2026-07-24
docs(logging): note the context logger de-duplicates logs across step replays

kang8:
The guide already says a log outside a `step` "can be run three times" and that the built-in wrapper mitigates it — but not that this depends on using the handler-provided `logger`. A logger you import or instantiate yourself and call directly outside a `step` still duplicates on every replay. This PR adds one sentence making that explicit, with the fix (use the handler `logger`, or wrap the log in `step.run`).

Docs only, no code change.

Backing (SDK source): `ProxyLogger` gates every method on an `enabled` flag (default false), enabled only in `onMemoizationEnd` — so it's silenced while memoized steps replay. See `packages/inngest/src/middleware/logger.ts` and `packages/inngest/src/components/Inngest.ts`.


kang8:
> However, the diff also removes the `metaTitle` and `description` exports

Hallucination?

PR #1806, opened 2026-07-24
docs(otel): tweak otel install guide

Linell:
Add install command and framework dev server preload to OpenTelemetry guide.


scottnuma (approved):
Looks good - gester also seems to have flagged a valid point

PR #1808, opened 2026-07-24
Add "Durable token streaming" docs pattern with inline CDN media

mitchellalderson:
Adds a new pattern page at /docs/patterns/durable/durable-token-streaming and teaches the patterns Markdown renderer to embed inline media.

- MarkdownRender: support standalone `alt` lines as a media block — images render as click-to-zoom (react-medium-image-zoom, same as docs), and `.mp4`/`.m3u8` URLs render as muted autoplay video (HLS via hls.js). Optional poster via `"poster=..."` in the Markdown title slot.
- Promote HlsVideo to shared/ so blog and patterns share one HLS player; add shared/Patterns/PatternMedia for the inline image/video branch.
- Register the pattern in patternsData (nav, listings, paths, prev/next, and agent/markdown views all generate automatically).

Uses plain /<video> for CDN media, so no next.config remotePatterns change is needed. The YouTube `video:` frontmatter slot is unchanged.


shared/Patterns/_patterns/durable-token-streaming.mdx
```diff
@@ -0,0 +1,269 @@
...
+  const flush = () => {
+    if (!buffer) return;
+    const delta = buffer;
+    buffer = "";
+    timer = null;
+    // Fire-and-forget, non-durable: a live UI nicety, not part of the
+    // durable record. `turn.completed` (below, durable) is the authoritative
+    // text the UI falls back to if any deltas are lost — so a failed publish
+    // is swallowed rather than left as an unhandled rejection.
+    void inngest.realtime.publish(ch.tokens, { turn, seq: seq++, delta }).catch(() => {});
```
djfarrelly: suggestion (non-blocking): I don't see folks using the `void` pattern very much, I get it's use, just may be a cleaner example code without it

djfarrelly (approved):

shared/Patterns/_patterns/durable-token-streaming.mdx
mitchellalderson: done, hardened it into a group of promises to e awaited in `finally()`

PR #1817, opened 2026-07-28
Fix: add redirects for markdown docs

djfarrelly:
Our redirects do not also handle the markdown docs like `docs-markdown` so some of those URLs dead end. This takes all the redirects and also copies them relative to the `docs-markdown` URLs as well so LLMs/scrapers aren't getting 404s.

Source: I pulled log data over a 24h period and saw lots of these.


patrickholcomb9 (approved):
nice flag & fix, Dan - sorry for missing these md redirects! lgtm!

PR #1830, opened 2026-08-06
Update and reorg sessions page with propagation

scottnuma:
- We're releasing propagated sessions, let's include it in the sessions docs
- Let's also reorg the sessions page to make it friendlier to get started. 
- I recommend reviewing by reading the [rendered sessions page as a whole](https://website-git-scott-propagated-sessions-doc-inngest.vercel.app/docs/features/events-triggers/sessions#top)
- EXE-2007
- Do not merge until sessions propagation is released with opt-out

PR #1831, opened 2026-08-06
Add Propagated Sessions Changelog

scottnuma:
- Add a change log for propagated sessions
- Before merging, we should update the date and we must have merged the docs updates: https://github.com/inngest/website/pull/1830
- EXE-2008

### DIFF
--- a/content/changelog/2026-08-07-propagated-sessions.mdx
+++ b/content/changelog/2026-08-07-propagated-sessions.mdx
@@ -1,4 +1,4 @@
-export const title = "Propagated sessions";
+export const title = "Propagated Sessions";
 export const date = "2026-08-07";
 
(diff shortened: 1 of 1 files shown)


content/changelog/2026-08-07-propagated-sessions.mdx
```diff
@@ -0,0 +1,8 @@
+export const title = "Propagated sessions";
+export const date = "2026-08-07";
+
+A run with sessions will now propagate its sessions into runs that it kicks off. Set sessions on just your root run, and then easily see the root run and all child runs together on the Sessions page.
```
amh4r: Something about this is kinda hard to parse mentally. What are your thoughts on this:

> Sessions now propagate to child runs. Any run triggered by another run (via event, invoke, or defer) inherits its parent's sessions. Set sessions once on the root run and the whole tree is grouped on the Sessions page.

scottnuma: I like this better, will use.

amh4r (approved):

PR #1837, opened 2026-08-10
docs(otel): add aiMiddleware quick start

Linell:
Add a quick start for `aiMiddleware()` to the OpenTelemetry guide, gated behind `?unreleased=ai-middleware` until the SDK feature ships, plus notes mapping its `traces` option to the Extended Traces behaviour modes.

SDK: inngest/inngest-js#1695 (aiMiddleware), inngest/inngest-js#1696 (example app).

### DIFF
--- a/content/changelog/2026-08-11-ai-middleware.mdx
+++ b/content/changelog/2026-08-11-ai-middleware.mdx
@@ -1,0 +1,25 @@
+export const title = "aiMiddleware: scoring, metadata, and Extended Traces in one bundle";
+export const date = "2026-08-11";
+export const unreleased = "ai-middleware";
+
+`aiMiddleware()` enables the TypeScript SDK's experimental AI features in one place: scoring (`step.score()`), step metadata (`step.metadata()`), and Extended Traces (`ctx.tracer`). Each previously needed its own middleware from `inngest/experimental`; one call now covers all three.
+
+```ts
+import { Inngest } from "inngest";
+import { aiMiddleware } from "inngest/experimental";
+
+export const inngest = new Inngest({
+  id: "my-app",
+  middleware: [...aiMiddleware()],
+});
+```
+
+Extended Traces attach to your registered OpenTelemetry provider by default; pass `traces: { behaviour: "off" }` to leave your provider untouched. Add your own middleware alongside the spread: `middleware: [...aiMiddleware(), myMiddleware]`.
+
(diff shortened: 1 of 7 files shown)


pages/docs/examples/open-telemetry.mdx
```diff
@@ -197,6 +227,12 @@ See the [`aiMetadata` client option](/docs/reference/typescript/v4/client/create
 
 Extended Traces export application OpenTelemetry spans from inside an Inngest function run to [Inngest Traces](/docs/platform/monitor/traces#extended-traces). Use this when you want HTTP calls, database queries, AI requests, and custom spans to appear in the run timeline.
 
+<Unreleased label="ai-middleware">
+
+`aiMiddleware()` from the [quick start](#quick-start) already enables Extended Traces. Use `extendedTracesMiddleware()` directly when you want Extended Traces without the rest of the bundle.
```
scottnuma, line 232: Can we warn people of the consequences of running both here too?

From the changelog
> In particular, running `extendedTracesMiddleware()` and `aiMiddleware()` together attaches two span processors to your provider, so every span is exported twice.

content/changelog/2026-08-11-ai-middleware.mdx
```diff
@@ -0,0 +1,25 @@
...
+
+export const inngest = new Inngest({
+  id: "my-app",
+  middleware: [...aiMiddleware()],
+});
+```
+
+Extended Traces attach to your registered OpenTelemetry provider by default; pass `traces: { behaviour: "off" }` to leave your provider untouched. Add your own middleware alongside the spread: `middleware: [...aiMiddleware(), myMiddleware]`.
+
+If you already use `scoreMiddleware()`, `metadataMiddleware()`, or `extendedTracesMiddleware()`, replace them with the bundle rather than adding it alongside them. In particular, running `extendedTracesMiddleware()` and `aiMiddleware()` together attaches two span processors to your provider, so every span is exported twice. Options you passed to `extendedTracesMiddleware()` move to the bundle's `traces` option: `extendedTracesMiddleware({ behaviour: "off" })` becomes `aiMiddleware({ traces: { behaviour: "off" } })`. The deprecated `"auto"` and `"createProvider"` behaviours are not part of the bundle; use [`@inngest/otel`](https://www.npmjs.com/package/@inngest/otel) to set up a provider instead.
```
scottnuma, line 19: We go a little deep into the configuration and potential errors of aiMiddleware here. Do you think we could omit them here, and that users would learn of these details looking in the deeper docs pages?

scottnuma (commented):
With AI middleware, we're introducing a new way for users to enable these three features. 

How should users choose between aiMiddleware and the original method? If we're offering this new method as a more convenient way, should we place it as the default way in all these docs?

I might also be missing useful context here - feel free to point me to some or resolve in Slack

PR #1851, opened 2026-08-14
Refresh durable execution page for SEO and clearer lifecycle messaging.

LCraigie:
Add a definition-led hero, Temporal/queues comparison with a Compare CTA, lifecycle glyphs, scoring/experiments in observability, and tighter primitive copy.

@whatupjohnb i changed up the icons at the top... but they're not really on brand. I just wanted to get back to some more useful animations. Can you have a look?

### DIFF
--- a/app/(v1)/platform/durable-execution/page.tsx
+++ b/app/(v1)/platform/durable-execution/page.tsx
@@ -8,9 +8,9 @@ import deDotsData from "@/public/assets/v1/durable-execution-hero/dots.json";
 const DE_DOTS_JSON = JSON.stringify(deDotsData);
 
 export const metadata: Metadata = generateMetadata({
-  title: "Durable Execution Explained: Automatic Retries & Checkpoints",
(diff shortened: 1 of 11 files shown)


whatupjohnb (commented):
I really like the icons but they are a bit floaty right now. Can you add a container around them to anchor them to the page a bit as  well as increase the scale? I can also pull this down and make the changes if that's easier. 

Table is nice but could use some more contrast between us and others

patrickholcomb9:
@whatupjohnb @LCraigie :

1) Updated the icon set under "Your workflows will fail. Then what?" section - adding a container around them made the formatting wonky; i asked Claude to replicate the icon treatment from https://www.inngest.com/platform/flow-control.. pls lmk if this is better?

2) Added a green hue gradient to the "Inngest" section of the compare table so that our feature column stands out more against competitors. lmk if you have any feedback here!

patrickholcomb9 (approved):
lgtm!

PR #1879, opened 2026-08-28
Add "If it Doesn't Compile, It Doesn't Count" sales overview PDF

patrickholcomb9:
Adds the compressed sales overview PDF to public/assets/overviews/if-it-doesnt-compile.pdf (3.48MB, down from 6.3MB).

/assets/overviews/if-it-doesnt-compile.pdf


rolesdurham (approved):
lgtm! ty pat

PR #1881, opened 2026-09-01
Add The AI Conference sponsorship to /events

rolesdurham:
Adds our sponsorship of **The AI Conference** (Pier 48, San Francisco, September 30 – October 1, 2026, booth #136) to the events page.

## What changed

| File | Change |
|---|---|
| `components/v1/sections/Events/data.ts` | Event entry in `UPCOMING` + `ALL_EVENTS`; new optional `endsAt` field on `EventItem` |
| `components/v1/pages/TheAIConference2026.tsx` | Detail page, with a "Register to attend" CTA → `https://aiconference.com/` |
| `app/(v1)/events/the-ai-conference-2026/page.tsx` | Route + metadata |
| `public/assets/v1/events/the-ai-conference-2026.png` | Card/hero graphic |

`/events` now shows an "Upcoming Events" section again — every event in the data had already started, so the section was correctly hiding itself.

## Reviewer notes

**⚠️ The graphic's dates don't match the copy.** The card artwork reads **"September 29 - October 1"**; the copy says **September 30 – October 1** (the dates I was given). Both are visible together in the hero, so one side needs to change before this ships — flagging for a decision rather than guessing. If Sept 29 is a pre-day/workshop and the expo floor opens the 30th, the copy could be reworded instead of re-cutting the graphic.

**New `endsAt` field.** `isPastEvent` derived past/upcoming from `startsAt` alone, so this multi-day conference would have flipped to "Past event" — losing its live Register CTA — at 9:01 AM on its own opening day. `endsAt` now takes precedence when present:

```ts
return new Date(ev.endsAt ?? ev.startsAt).getTime() < Date.now();
```

No existing event sets `endsAt`, so behavior is unchanged for all of them. Note the AI Engineer World's Fair entry has the same latent issue if it's ever re-run.

**Other upcoming events rail.** This detail page derives it from the shared `ALL_EVENTS` rather than the ~170-line hand-copied list every other event detail page carries. Those copies have already drifted — e.g. the AIEWF page still points August's {AI} in Production at Luma instead of the YouTube recording linked in #1880. Worth a follow-up to converge them; not done here to keep this PR scoped.

**`imageFit: "contain"`.** The artwork is text-heavy (booth number, dates, venue on the left panel) and `cover` cropped those lines off in the 32.5% card column. Matches the treatment already used for the other text-heavy event cards.

## Verification

- Typecheck clean (`tsc --noEmit`), prettier run
- Dev server checked at desktop and mobile (375×812): `/events` and `/events/the-ai-conference-2026` both render, no console errors
- "Other upcoming events" rail correctly hidden on the detail page (every other event is past)


patrickholcomb9 (approved):
lgtm

PR #1899, opened 2026-09-10
Fix The AI Conference card art and event ordering on /events

rolesdurham:
Three fixes to `/events`, all surfaced by the AI Conference entry. Split out of #1892, which had one of them committed alongside the long-run landing page work.

### The AI Conference card art

The entry was set to `imageFit: "contain"`, which letterboxed the full-bleed 16:9 artwork against the card's gradient. Plain `cover` fixed the letterbox but clipped the type on the left, because the featured card's image column is ~1.56 and crops the sides rather than the top and bottom.

It's now `cover` anchored left, so the crop lands on the bleed pattern on the right and the whole left panel survives. That needed a new optional `imagePosition` on `EventItem`, wired into both card components and defaulting to `center`, so no other card changes.

### Upcoming events sorted above past ones

`sortEventsByDate` returned `[...past, ...upcoming]`, so anything still open to register for sorted below every event that had already happened — the AI Conference sat at the very bottom of "All Events". Now `[...upcoming, ...past]`: upcoming soonest-first, then past newest-first.

`UpcomingEvents` was also rendering `UPCOMING` in raw array order, which looks right today only because the AI Conference is the sole non-past entry; the next event appended would render last regardless of date. It sorts now too.

The detail-page "Other upcoming events" rails filter to non-past before sorting, so they're unaffected.

### Conference vs. booth dates

The artwork reads "September 29 - October 1" because that's the conference; the site said "September 30 - October 1" because that's when our booth is staffed. Neither was wrong, so the metadata and page copy now state both instead of contradicting the graphic.

The displayed date and `startsAt`/`endsAt` stay on the booth dates, so the card keeps its live Register CTA through the last day we're there and flips to "past" only after.

### Also included

The `contain` letterbox fix for the event cards generally (`bg-black` behind `contain` images, and the Innhouse photo switched to `cover`). This was previously commit `9c455537` on #1892; that branch has been rewritten to contain only long-run changes, so the two PRs no longer share files.

### Verification

`tsc --noEmit` clean. `/events` and `/events/the-ai-conference-2026` checked in the dev server: card art fills with all type intact at both card sizes, AI Conference leads both sections, past events follow newest-first.

### DIFF
--- a/app/(v1)/events/the-ai-conference-2026/page.tsx
+++ b/app/(v1)/events/the-ai-conference-2026/page.tsx
@@ -5,7 +5,7 @@ import TheAIConference2026 from "@/components/v1/pages/TheAIConference2026";
 export const metadata: Metadata = generateMetadata({
   title: "Meet Inngest at The AI Conference 2026",
   description:
-    "Inngest is agent infrastructure that lives in your codebase. Find us at booth #136 at Pier 48 in San Francisco, September 30 – October 1, 2026.",
+    "Inngest is agent infrastructure that lives in your codebase. The AI Conference runs September 29 – October 1, 2026 at Pier 48 in San Francisco — find us at booth #136 on September 30 and October 1.",
 });
 
 export default function Page() {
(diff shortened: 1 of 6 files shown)


patrickholcomb9 (approved):
lgtm 🚀

PR #1912, opened 2026-09-18
docs(blog): incident report for September 18, 2026

mar-inngest:
## Summary

Public incident report for INC-1111 (Sept 18, 2026) — PgBouncer connection exhaustion caused by a Vercel Marketplace account hard-delete cascade and its ten retries.

Authors: Jakob Evangelista (lead), Muzammil A., Albert Chae, Bruno Scheufler.

## Sources

- INC-1111 (incident.io) + `#inc-2026-09-18-run-scheduling-volume-is-low`
- Jakob's log analysis: first deletion 15:39:39 UTC, 10 retries 15:41–15:56, cancellations 16:39, locks cleared ~16:42
- Datadog: PgBouncer client conns 427 → 6,066 (5,933 waiting), Postgres locks 544 → 9,269 @ 16:16 UTC
- Mitigations referenced: inngest/monorepo#8774 (`VERM_DELETIONS_DISABLED`), inngest/ansible#495 (PgBouncer config)

## Open items before un-drafting

- [ ] Incident is still open in incident.io — confirm final status and that no further impact occurred
- [ ] Two facts stated as unknown in the post: who initiated the original Vercel deletion, and why the first cascade ran so long. Fill in or keep the explicit "still investigating" wording
- [ ] Confirm the customer-impact window (16:18–16:42 UTC, recovered 16:51) is what we want to state publicly
- [ ] Confirm whether enterprise customers on dedicated infrastructure were affected — this is not yet answered and the post does not claim either way
- [ ] Decide whether to name Vercel Marketplace explicitly (current draft does; the deletion requests themselves were legitimate)


content/blog/2026-09-18-incident-report-for-september-18-2026.md
```diff
@@ -0,0 +1,95 @@
...
+  - Albert Chae
+  - Bruno Scheufler
+category: engineering
+---
+
+_All timestamps are in UTC._
+
+## Summary
+
+On September 18, 2026, a single account deletion originating from our Vercel Marketplace integration, retried ten times, blocked our primary Postgres database for long enough to exhaust the connection pool in front of it. From 16:18 to 16:42 UTC, both poolers were rejecting new connections, and services across the platform could not reach the database. This stalled run scheduling and event acknowledgement.
```
mar-inngest: say our poolers instead of both poolers

content/blog/2026-09-18-incident-report-for-september-18-2026.md
```diff
@@ -0,0 +1,111 @@
...
+  - Albert Chae
+  - Bruno Scheufler
+category: engineering
+---
+
+_All timestamps are in UTC._
+
+## Summary
+
+On September 18, 2026, a single account deletion originating from our Vercel Marketplace integration, retried ten times, blocked our primary Postgres database for long enough to exhaust the connection pool in front of it. Services across the platform could not reach the database, which stalled run scheduling and event acknowledgement.
```
mar-inngest: do we want to mention " retried ten times,"? 
Also, maybe remove "single account deletion" and say "account deletion"

content/blog/2026-09-18-incident-report-for-september-18-2026.md
```diff
@@ -0,0 +1,111 @@
...
+
+There were two periods of impact: 16:18 to 16:42, and a second, slightly larger one from 17:50 to 18:12 that occurred while we were rolling out the mitigation.
+
+Events sent to Inngest during the incident were durably accepted, and scheduling resumed once the blocking transactions cleared. Affected work was delayed rather than lost.
+
+## What Happened
+
+At 15:39:39, our Vercel Marketplace integration received an installation deletion and began deprovisioning the corresponding Inngest account. The request path was: Vercel installation deletion, to marketplace deprovisioning, to a cascading account delete.
+
+Vercel requires hard deletion rather than the soft deletion we use everywhere else in the product. Our deprovisioning path therefore deletes the user inside a transaction and then issues a real `DELETE` against the account row. That second statement cascades: accounts and workspaces are referenced by foreign keys from dozens of tables, most declared `ON DELETE CASCADE`. One account deletion fans out into deletes across a large portion of the schema, inside one transaction, holding locks on everything it touches until it commits.
```
mar-inngest: Did it really require hard deletion? If so that's fine.

Also, before this mention that we use postgres for account management (or something) and that we periodically query postgres for updates to customer information (something about changing tiers, moving customers between different pieces of the infrastructure, etc. etc)

content/blog/2026-09-18-incident-report-for-september-18-2026.md
```diff
@@ -0,0 +1,111 @@
...
+
+This first cascade ran far longer than expected. We have not yet established why; the historical logs do not preserve the complete database blocking chain.
```
mar-inngest: Do we need to mention this? "This first cascade ran far longer than expected. We have not yet established why; the historical logs do not preserve the complete database blocking chain."

content/blog/2026-09-18-incident-report-for-september-18-2026.md
```diff
@@ -0,0 +1,111 @@
...
+
+What turned a slow transaction into an outage was the retry behavior around it. Between 15:41 and 15:56, ten additional deletion attempts arrived for the same installation and the same account. Each one tried to delete rows the first transaction already held locks on, so instead of failing fast they queued behind it. One user-delete query ended up waiting 44 minutes. Lock counts on the primary climbed from a baseline of roughly 540 to a peak of 9,269 at 16:16, and active backend connections rose from about 180 to just over 1,000.
+
+Blocked queries do not release their connection. Each one continued to occupy a client slot in PgBouncer for the entire wait, and the pressure spread to every service sharing that pool, including services with no relationship to the deletion. PgBouncer client connections went from a steady 427 to 6,066, with 5,933 of them waiting for a server connection. From 16:09 the pooler queues grew steadily, and by 16:18–16:19 both PgBouncer hosts had hit their 3,000 client limit. They began rejecting new connections and disconnecting waiting clients at the 120 second wait timeout:
```
mar-inngest: don't mention the 3,000 client limit. don't mention 427 to 6,066, with 5,933, but mention multiplication factor

content/blog/2026-09-18-incident-report-for-september-18-2026.md
mar-inngest: Done — "our poolers" throughout, including the timeline entries.

content/blog/2026-09-18-incident-report-for-september-18-2026.md
mar-inngest: Done. Summary now reads "an account deletion" with no retry count. The retry pile-up is still in **What Happened** and as root-cause factor 2, since it's what turned one slow transaction into an outage — happy to trim it there too if you'd rather it not be customer-facing.

content/blog/2026-09-18-incident-report-for-september-18-2026.md
mar-inngest: Good catch — I can't substantiate that Vercel mandates it. The only evidence is our own code comment ("Vercel requires hard deletion") in `pkg/marketplace/deprovision.go`, and Albert's ghost-account proposal implies it isn't a hard requirement. Reworded to describe our implementation and its motivation (nothing left behind that blocks a reinstall) rather than asserting a Vercel constraint.

Also added two paragraphs before it: what Postgres does for us (accounts, workspaces, API keys; entitlements and tier changes, shard placement, config changes, queried continuously) and that every service reaches it through shared PgBouncer pools.

content/blog/2026-09-18-incident-report-for-september-18-2026.md
mar-inngest: Removed from that spot. The same acknowledgement still appears once at the end of **Root Cause** — that we haven't established who initiated the deletion or why the first cascade ran so long. Given we're publishing mid-investigation I'd keep that one, but say the word and it goes.

content/blog/2026-09-18-incident-report-for-september-18-2026.md
mar-inngest: Done. No client limit, no raw connection counts. Now: client connections "roughly fourteen times their steady-state level", locks "roughly seventeen times baseline", backends "about five times normal". I applied the same treatment to the second event ("marginally worse than the first peak") and to the 720-backend figure, on the assumption you don't want capacity numbers published anywhere.

jakobevangelista (approved):

content/blog/2026-09-18-incident-report-for-september-18-2026.md
```diff
@@ -0,0 +1,100 @@
...
+focus: false
+featured: false
+heading: "Incident report for September 18, 2026 - When Deletion Cascades Exhausted Our Database Connections"
+subtitle: A report on the database connection pool exhaustion that interrupted run scheduling and event processing.
+image: /assets/blog/incident-post-mortem.svg
+date: 2026-09-18
+author:
+  - Jakob Evangelista
+  - Muzammil A.
+  - Albert Chae
```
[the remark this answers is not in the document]
jakobevangelista: ^^^ nit, it would be cool if gester could suggest the code change 🤔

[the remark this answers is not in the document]
mar-inngest: Good catch — fixed in ec221984. The second window was shorter (22 min vs 24 min) but hit more services, so the summary now says it "reached more services than the first" rather than calling it larger.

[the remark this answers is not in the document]
mar-inngest: Sorry, Jakob, these were two AIs 🤦

albertchae (approved):

